PR #2, opened 2016-05-20
add missing textdomains for translations

kniebremser:


helen:
Hi @kniebremser! I was reviewing the plugin to make sure it can work with language packs and noticed this PR. I'll fix up the last one and (hopefully) give you co-author credit here in GH. The strings about ratings are from core and don't need the domain. Closing because this is stale.

PR #7, opened 2018-02-15
Multisite / Network Optimizations for $this->get_avatar()

tliebig:
I've made some minor changes to $this->get_avatar because it was not pulling/creating correctly resized images when requesting a specific size.

For changes please see the attached simple-local-avatars.php
[simple-local-avatars.php.zip](https://github.com/10up/simple-local-avatars/files/1728420/simple-local-avatars.php.zip)

P.S.: Im new to git, and pretty sure I messed up something since I'm reading "6 commits" where there should be only one. The attached file is the one that includes all the changes to simple-local-avatars.php. If there's something I can do to clean up, please tell me.


helen:
Hi there - I'm so sorry to have neglected this for so long. Because there have been changes in that file since you uploaded the ZIP, it's somewhat difficult to figure out what exactly your proposed changes are. Could you possibly try opening a pull request with changes from your own fork - one way to do this is by clicking the "Edit this file" button in GitHub and making your changes there, then submitting your suggestion for changes. Thanks!

tliebig:
Hi Helen,
I've just edited the via via the edit browser button. The changes were added to v2.1., I've upgraded my local version to that yesterday and re-inserted the changes afterwards. I've modified how the path to the image is calculated, and added a simple cache via site_transient. The diff shows what has been added, which should be quite overseeable :o)

PR #8, opened 2018-07-19
use media library interface to upload files

eflorea:
remove the upload button and functionality since this is already provided by the media library popup interface, avoid conflicts with other plugins like Windows Azure Storage or AWS S3.

### DIFF
--- a/simple-local-avatars.php
+++ b/simple-local-avatars.php
@@ -252,11 +252,6 @@ class Simple_Local_Avatars {
 						'_wpnonce'	=> $this->remove_nonce,
 					) );
 			?>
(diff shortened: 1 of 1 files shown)


helen:
This looks fine to me code-wise. UI-wise, I think the buttons now look kind of strange next to the image instead of under it since there's nothing else accompanying them. We can do that in a separate PR though. I'll merge this once I figure out what the best thing to do about the extraneous commits in develop is :)

helen:
In reviewing this further for .org release, I realized that the Media Library is only available for people with the `upload_files` capability and this PR effectively disabled any uploads for subscribers :( Will have to think about a different approach here.

PR #9, opened 2018-08-16
add composer.json file

eflorea:
Add composer.json so plugin can be installed via composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "wordpress",
     "10up"
   ],
-  "homepage": "https://github.com/10up/simple-local-avatars"",
+  "homepage": "https://github.com/10up/simple-local-avatars",
   "license": ["GPL-2.0-only"],
   "authors": [
     {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -0,0 +1,30 @@
...
+    }
+  ],
+  "support": {
+    "issues": "https://github.com/10up/simple-local-avatars",
+    "source": "https://github.com/10up/simple-local-avatars"
+  },
+  "require": {
+    "php": ">=5.3"
+  },
+  "autoload": {
```
helen: Based on https://github.com/10up/10up-experience/pull/22/files it seems like maybe we shouldn't use autoload here?

composer.json
```diff
@@ -0,0 +1,30 @@
...
+  "homepage": "https://github.com/10up/simple-local-avatars",
+  "license": ["GPL-2.0-only"],
+  "authors": [
+    {
+      "name": "10up",
+      "homepage": "https://10up.com/"
+    }
+  ],
+  "support": {
+    "issues": "https://github.com/10up/simple-local-avatars",
```
helen: I imagine this should be https://github.com/10up/simple-local-avatars/issues

PR #10, opened 2018-08-31
Add support for getting and updating via the REST API

robbiet480:
Adds support for getting and updating the avatar via the REST API. Only providing the `media_id` is supported as you obviously can't change the URLs.

### DIFF
--- a/simple-local-avatars.php
+++ b/simple-local-avatars.php
@@ -48,6 +48,8 @@ class Simple_Local_Avatars {
 		add_action( 'user_edit_form_tag', array( $this, 'user_edit_form_tag' ) );
 		
 		add_filter( 'avatar_defaults', array( $this, 'avatar_defaults' ) );
+
+		add_action( 'rest_api_init', array( $this, 'register_rest_fields' ) );
 	}
 
 	/**
@@ -472,6 +474,48 @@ class Simple_Local_Avatars {
 	public function user_profile_update_errors( WP_Error $errors ) {
 		$errors->add( 'avatar_error', $this->avatar_upload_error );
 	}
+
+	/**
+	 * Registers the simple_local_avatar field in the REST API.
+	 */
+	public function register_rest_fields() {
+		register_rest_field( 'user', 'simple_local_avatar', array(
+			'get_callback' => array( $this, 'get_avatar_rest' ),
+			'update_callback' => array( $this, 'set_avatar_rest' ),
+			'schema' => array(
+				'description' => 'The users simple local avatar',
+				'type' => 'object',
+			)
+		));
+	}
+	
+	/**
+	 * Returns the simple_local_avatar meta key for the given user.
+	 * 
+	 * @param object $user User object
+	 */
+	public function get_avatar_rest( $user ) {
+		$local_avatar = get_user_meta( $user['id'], 'simple_local_avatar', true );
+		if ( empty( $local_avatar ) ) {
+			return;
+		}
+		return $local_avatar;
+	}
+
+	/**
+	 * Updates the simple local avatar from a REST request.
+	 *
+	 * Since we are just adding a field to the existing user endpoint
+	 * we don't need to worry about ensuring the calling user has proper permissions.
+	 * Only the user or an administrator would be able to change the avatar.
+	 * 
+	 * @param array $input Input submitted via REST request.
+	 * @param object $user The user making the request.
+	 */
+	public function set_avatar_rest( $input, $user ) {
+		$this->assign_new_user_avatar($input['media_id'], $user->ID);
+	}
+
 }
 
 $simple_local_avatars = new Simple_Local_Avatars;


simple-local-avatars.php
```diff
@@ -1,571 +1,610 @@
...
+			'update_callback' => array( $this, 'set_avatar_rest' ),
+			'schema' => array(
+				'description' => 'The users simple local avatar',
+				'type' => 'object',
+			)
+		));
+	}
+
+	/**
+	 * Returns the simple_local_avatar meta key for the given user
```
helen: This docblock needs some ending periods. :)

simple-local-avatars.php
```diff
@@ -1,571 +1,610 @@
...
+	 * @param object $profileuser User object
```
helen: I'm guessing this param is supposed to be `$user`? And the one below a leftover from something else that isn't needed anymore.

simple-local-avatars.php
```diff
@@ -1,571 +1,610 @@
...
+	 * @param array|string $input Passed input values to sanitize
+	 */
+	public function get_avatar_rest( $user ) {
+		$local_avatar = get_user_meta( $user[ 'id' ], 'simple_local_avatar', true );
```
helen: Mild whitespace tweak - remove from around `'id'` and add inside `empty()` on the line below.

simple-local-avatars.php
```diff
@@ -1,571 +1,610 @@
...
+		if ( empty($local_avatar) ) {
+			return;
+		}
+		return $local_avatar;
+	}
+
+	/**
+	 * Updates the simple local avatar from a REST request.
```
helen: I think it would be good to note here in the doc block what the built-in REST permissions check is so that in the future we don't wonder why we aren't worried about checking user caps or nonces. Or should we be worried about that?

helen (changes requested):
Your final commit seems to have changed all the line endings, and there are some other whitespace changes sprinkled in here and there, I imagine by your editor. Could you possibly revert all whitespace changes that aren't on lines you've added? We can fix existing trailing whitespace and CRLF issues separately, I am just a bit of a stickler for a readable history (will likely squash this particular merge).

helen:
Thanks for the PR! I left some code comments in a review :) I have another question about the general idea as well - would it make sense to return the URL for the avatar as well? I guess it's a matter of whether we do that additional query on the server side or the requester has to do a separate request for the actual media item?

robbiet480:
@helen My thinking was that since you can only pass in a media ID to update the avatar that we should only expose the media ID. Happy to implement returning URLs in the API if you think that would be better though!

robbiet480:
@helen I _think_ I got the line endings and whitespace issues taken care of. I force pushed a single commit with all my fixes (normally I'd just push another commit but thought between all the whitespace/line ending issues and the need-to-fix comments was better to just start with a clean slate).

helen:
@robbiet480 Makes sense re: the media ID. Easier to add later than have to remove for some reason, in any case. The diff looks good now, I'll review and test this and get it merged.

helen (approved):

simple-local-avatars.php
```diff
@@ -472,6 +474,48 @@ public function unique_filename_callback( $dir, $name, $ext ) {
...
+		));
+	}
+	
+	/**
+	 * Returns the simple_local_avatar meta key for the given user.
+	 * 
+	 * @param object $user User object
+	 */
+	public function get_avatar_rest( $user ) {
+		$local_avatar = get_user_meta( $user['id'], 'simple_local_avatar', true );
```
helen: So this does seem to be working, but I would have thought it would need to be `$user->ID` since it's (allegedly) an object?

helen: Ah I see, so it is actually an array during this process. The docblock should reflect that :)

t-lock:
Been banging my head against this for awhile, since I'm a bit of a rest api noob (somehow got into GraphQL first for client-side apps and just did WP development the old way).

Figured I'd drop a comment in case it can help anyone who discovers this PR via google (I don't see any official documentation for this plugin, or anything in the support forums on how to edit avatar via rest api)

```
    const id = 666; // your media id, likely as a return from hitting a media upload endpoint
    const url = `${window.location.origin}/wp-json/wp/v2/users/me/`; // substitute user id for "me" if desired 
    fetch(url, {
      method: "PATCH",
      headers: {
        // pass a nonce from php to js via wp_localize_script() and wp_create_nonce('wp_rest')
        "X-WP-Nonce": wpVars.nonce,
        "Content-Type": "application/json",
      },
      body: JSON.stringify({ simple_local_avatar: { media_id: id } }),
    })
      .then((response) => response.json())
      .then((data) => console.log(data));
```

@helen @robbiet480 if you are still part of the core team here, maybe this support ticket could use a little love: https://wordpress.org/support/topic/how-can-i-change-users-avatar-via-rest-api/  

Thanks for your work on this plugin!! :green_heart:

PR #14, opened 2018-09-13
add compatibility with plugins to upload images to Cloud storage (Azure Storage, AWS S3)

eflorea:
use media_handle_upload instead of wp_handle_upload, update plugin version


helen:
I pushed a commit to revert out the whitespace changes so we can evaluate what's really changed here and for a cleaner history. Now that I am looking at the actual changes - I think this means that non-privileged users who upload avatars are now creating attachments rather than just having some files floating around the upload directory. What does that end up meaning for the "author" of the attachment (i.e. does having an otherwise non-privileged user assigned as the author of that attachment have any side effects), and does this potentially create a bunch of clutter in the media library? I guess on the upside you can now much more easily delete a bunch of avatars at once without having to go into each individual profile or FTP in (though the reference will still remain in user meta) but looking at the `avatar_delete()` method, those attachments won't get deleted later whereas the current way of uploaded non-attachments do get deleted ("unlink"-ed).

PR #19, opened 2019-02-28
add args support to get_avatar

heyjones:
Our $args were getting dropped, so I just modified the plugin to accept that array.

### DIFF
--- a/simple-local-avatars.php
+++ b/simple-local-avatars.php
@@ -3,7 +3,7 @@
  * Plugin Name: Simple Local Avatars
  * Plugin URI: https://10up.com/plugins/simple-local-avatars-wordpress/
  * Description: Adds an avatar upload field to user profiles. Generates requested sizes on demand, just like Gravatar! Simple and lightweight.
- * Version: 2.1
+ * Version: 2.1.1
  * Author: Jake Goldman, 10up
  * Author URI: https://10up.com
  * License: GPLv2 or later
@@ -32,7 +32,7 @@ class Simple_Local_Avatars {
 
 		// supplement remote avatars, but not if inside "local only" mode
 		if ( empty( $this->options['only'] ) )
-			add_filter( 'get_avatar', array( $this, 'get_avatar' ), 10, 5 );
+			add_filter( 'get_avatar', array( $this, 'get_avatar' ), 10, 6 );
 		
 		add_action( 'admin_init', array( $this, 'admin_init' ) );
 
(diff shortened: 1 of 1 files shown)


jeffpaul:
@heyjones thanks for the PR, I'm adding this to our next milestone release (2.2) but that's yet to be planned so no timeline yet on when that would be available.

Pinging @helen @eflorea for review.

simple-local-avatars.php
```diff
@@ -140,6 +140,9 @@ public function get_avatar( $avatar = '', $id_or_email = '', $size = 96, $defaul
 			$local_avatars[$size] = home_url( $local_avatars[$size] );
 		
 		$author_class = is_author( $user_id ) ? ' current-author' : '' ;
+		if( array_key_exists( 'class', $args ) ){
+			$author_class .= ' ' . $args['class'];
```
helen: I'm guessing this is here to mimic what core does with `$args['class']`; however, I'm not sure we want to get into adding it in this particular PR without fully considering the other expected args.

heyjones (commented):
Should I submit an updated PR that handles all of the arguments? I think we’d only need args that manipulate the img tag output...

adamsilverstein:
What args does core support here? I don't see 'class' here - https://developer.wordpress.org/reference/functions/get_avatar_data/

heyjones:
My guess is that it gets handled in the 'extra_attr' arg?

simple-local-avatars.php
heyjones: A full list of args is available here:

https://codex.wordpress.org/Function_Reference/get_avatar

I can make some modifications to support the other properties in $args (size, height, width, etc) and resubmit, though that will take some pretty significant changes to how the <img tag is being rendered.

heyjones: I'll just try duplicating the code from the native `get_avatar()` function:

https://core.trac.wordpress.org/browser/tags/5.1.1/src/wp-includes/pluggable.php#L2546

adamsilverstein: Hey @heyjones thanks for your work here, this looks great! 

Mimicking what `get_avatar` does makes sense, do you think you will be able to finish this up or would you like some help here to get that added?

heyjones: I'm updating this PR to reflect your 2.1.1 update, going to try to get those additional args implemented.

heyjones (commented):
I started to attempt to sync your get_avatar override with core and add support for all of the $args, but I suspect that might introduce breaking changes (ie: `.current-author`):

https://core.trac.wordpress.org/browser/tags/5.2/src/wp-includes/pluggable.php#L2546

dinhtungdu (commented):
Thank @heyjones, this PR works. But as mentioned in the conversation, we're mimicking the changes in core here, it means that whenever the core `get_avatar` is updated, we need to follow that change.
Can you take a look at #40 which completely removes the overridden `get_avatar` function, it may be a better alternative to this?

jeffpaul:
Closing in favor of #40

PR #34, opened 2019-10-11
Explicitly declare global

pdewouters:
### Description of the Change

Fixes an error related to variable scope

`Uncaught Error: Call to a member function get_avatar() on null in /chassis/vendor/10up/simple-local-avatars/simple-local-avatars.php on line 580get_simple_local_avatar()vendor/10up/simple-local-avatars/simple-local-avatars.php:254edit_user_profile()wp-includes/class-wp-hook.php:286apply_filters()wp-includes/class-wp-hook.php:310do_action()wp-includes/plugin.php:465do_action()wp-admin/user-edit.php:700require_once()wp-admin/profile.php:18
`

### Benefits

The plugin can be included via our loading mechanism

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/simple-local-avatars/blob/develop/CONTRIBUTING.md) document.

### Changelog Entry
Initialize `Simple_Local_Avatars` on the `$simple_local_avatars ` global, enabling bundling plugin with composer.

### DIFF
--- a/simple-local-avatars.php
+++ b/simple-local-avatars.php
@@ -564,7 +564,8 @@ class Simple_Local_Avatars {
 
 }
 
-$GLOBALS['simple_local_avatars'] = new Simple_Local_Avatars();
+global $simple_local_avatars;
+$simple_local_avatars = new Simple_Local_Avatars();
 
 /**
  * more efficient to call simple local avatar directly in theme and avoid gravatar setup
--- a/simple-local-avatars.php
+++ b/simple-local-avatars.php
@@ -564,7 +564,7 @@ class Simple_Local_Avatars {
 
 }
 
-$simple_local_avatars = new Simple_Local_Avatars;
+$GLOBALS['simple_local_avatars'] = new Simple_Local_Avatars();
 
 /**
  * more efficient to call simple local avatar directly in theme and avoid gravatar setup


adamsilverstein (changes requested):
Hi @pdewouters and thanks for the PR!

Can you give some more details about why you want to make this change? 

* What exact conditions are used to create the error you show above, can you share your loading code?
* Have you considered other approaches? 
* Could this change have unintended side effects?

pdewouters:
@adamsilverstein yes here's where the plugin is loaded: https://github.com/humanmade/altis-cms/pull/111/files , we want to include it via composer, and then load it on the `plugins_loaded` action by including the plugin file manually.
The reason this error occurs is because `$simple_local_avatars` is not in the global scope when the plugin file is included via a calling function.

Other approach could be:

```
global $simple_local_avatars;
$simple_local_avatars = new Simple_Local_Avatars();
```

Instead of making it a super global with `$GLOBALS` - this might be preferable?

adamsilverstein:
@pdewouters thanks for clarifying, i'll take a closer look and let you know what we can do here.

adamsilverstein:
@pdewouters - this looks good, i verified the issue and the fix. 

can you please change to your suggested 
```
global $simple_local_avatars;
$simple_local_avatars = new Simple_Local_Avatars();
```
to better match the rest of the code

simple-local-avatars.php
```diff
@@ -564,7 +564,7 @@ public function set_avatar_rest( $input, $user ) {
 
 }
 
-$simple_local_avatars = new Simple_Local_Avatars;
+$GLOBALS['simple_local_avatars'] = new Simple_Local_Avatars();
```
adamsilverstein:
```suggestion
global $simple_local_avatars;
$simple_local_avatars = new Simple_Local_Avatars();
```

adamsilverstein (approved):
Thanks @pdewouters!

pdewouters:
Thanks for merging @adamsilverstein , any plans for a new tagged release?

jeffpaul:
@pdewouters per https://github.com/10up/simple-local-avatars/issues/51, we're looking to release 2.2.0 sometime in July.  We've got 2 releases ahead of it on other open source projects, but hope to get to 2.2.0 soon.

PR #40, opened 2020-01-20
Fix: Adapt to More Current Avatar Filters

### LINKED ISSUE #39: Adapt to More Current Avatar Filters
Simple Local Avatars hooks into the get_avatar filter. However, there are better options introduced after initial creation of this plugin. 

Introduced in 4.2.0, get_avatar_data returns an array of arguments. This seems better than the #19 solution. 

The pre_get_avatar filter seems best. It can check for the presence of a user_id or an object with one, and if not return control to other code. 

It is more involved to implement, but it would make for a massive improvement and would support things like srcset, which the plugin currently overrides and ignores. It is also better for future proofing. It would allow the removal of redundant code as well.

Leaving it as is?

**Additional context**
The minimum requirement is WordPress 4.6, so there is no compatibility issue

### LINKED ISSUE #49: Incompatibility with Elementor
**Describe the bug**
[Previously reported on WordPress.org](https://wordpress.org/support/topic/it-doesnt-work-with-elementor-pro/#post-12911947):
> It doesn’t work with Elementor PRO?

**Steps to Reproduce**
First report:
> I tested the plugin on an author box built with Elementor, but the Gravatar is not shown.
It would be good to check why, due to the popularity of Elementor.

Second report:
> I’m facing the same issue with elementor pro. I’m trying to use Simple Local Avatars for my profile avatar then I tried to use it with the dynamic content available in elementor pro, but the profile picture doesn’t show up when I select the profile picture in the dynamic section. 

**Expected behavior**

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - Simple Local Avatars version: 
 - Theme and version: 
 - Other installed plugin(s) and version(s): WordPress 5.4.1 and Elementor Pro 2.9.5

**Additional context**

### LINKED ISSUE #57: Implement `get_avatar_url` the same as `get_avatar`
Yes, currently you can get the avatar from simple local avatars by calling `get_avatar` and passing the appropriate arguments. We would like to be able to do the same thing with `get_avatar_url`. Specifically in our case the issue is that the WPGraphQL plugin uses the `get_avatar_url` to populate the `avatar` field when requesting user information. When this happens we only get back the default Gravatar icons and not the simple local avatar that we have created.

We would like to be able to use the simple local avatars with our GraphQL and in order to do that we need to have `get_avatar_url` use simple local avatars as well

We could also override the `get_avatar_url` filter in our functions file of the theme. We'd prefer to not do that since that ties our theme very tightly to this plugin. It also feels more correct to have the plugin return the URL when calling `get_avatar_url`

**Additional context**
Additionally if `get_avatar_url` returns the url then the `get_avatar` function could call that to get the appropriate URL to add to the `src`

dinhtungdu:
### Description of the Change

* Switch to `pre_get_avatar_data` filter.
* Remove `get_avatar` function that overrides the core one.
* Add `$args` parameter to `get_simple_local_avatar` function.
* ~Remove `Simple_Local_Avatars::get_avatar` method.~ ~> added back for backward compat.
* Add `Simple_Local_Avatars::get_avatar_data()` method.
* Add `Simple_Local_Avatars::get_simple_local_avatar_url()` method.
* Add `Simple_Local_Avatars::get_default_avatar_url()` method.
* Fix coding standard issues.

### Benefits

In fact, we only need to filter the url of the avatar, not the avatar
img element. So instead of hooking to `get_avatar` filter and overriding
the `get_avatar` function, we hook to `pre_get_avatar_data` filter to
manipulate the avatar url and let the core handle the rest. This reduces
duplication code and minimizes effort to follow core changes.

### Possible Drawbacks

We need to care about/check for backward compatibility issues.

### Verification Process

Check if avatars display correctly when `Local Avatars Only` is enabled/disabled.

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/simple-local-avatars/blob/develop/CONTRIBUTING.md) document.

### Applicable Issues
Closes #39
Relates #19


jeffpaul:
Pinging @dkotter and @ryanwelcher on this to see if they can review this PR.

jeffpaul:
@heyjones - I wanted to check and see if you'd mind giving this PR a review in relation to your PR to see if this approach might also work for your use?

heyjones:
Sorry for the delay... it maintains the class args I'm passing through the theme!

@dinhtungdu I'd recommend possibly running `$this->options` through `wp_parse_args()` as I had a PHP notice locally since I didn't have the `only` option set:

`Notice: Undefined index: only ... on line 70`

Aside from that, @jeffpaul as long as you don't see this causing any issues with backwards capability I think it's a really clever solution!

dinhtungdu:
@heyjones, thanks for the suggestion, I fixed it in https://github.com/10up/simple-local-avatars/pull/40/commits/00a83620a6d1f1bc43d54a6d921ba4d3fa90d037

MikeNGarrett:
I really like this method better than continuing support via a pluggable function.

lincolnlemos:
This is what I was looking for! Thanks 👍

MikeNGarrett:
Is there an expected timeline for when this PR will be released?

jeffpaul:
@MikeNGarrett one item I'd like to see is some test coverage, be that unit tests and/or WP Acceptance tests, before pushing forward with another release.  If someone is able to add a PR to increase (heck, add) test coverage then that'd be wonderful.

simple-local-avatars.php
```diff
@@ -134,11 +152,38 @@ public function get_avatar( $avatar = '', $id_or_email = '', $size = 96, $defaul
...
+	 * Get default avatar url 
+	 *
+	 * @since 2.2.0
+	 *
+	 * @param int $size Requested avatar size. 
+	 */
+	public function get_default_avatar_url( $size ) {
+		if ( empty( $default ) ) {
+			$avatar_default = get_option( 'avatar_default' );
+			if ( empty( $avatar_default ) )
```
dkotter: I know this was just copied from the current codebase but we should make sure any if statements here have brackets `{ }` at the end of the lines.

simple-local-avatars.php
```diff
@@ -134,11 +152,38 @@ public function get_avatar( $avatar = '', $id_or_email = '', $size = 96, $defaul
...
+				$default = 'mystery';
+			else
+				$default = $avatar_default;
+		}
+
+		$host = is_ssl() ? 'https://secure.gravatar.com' : 'http://0.gravatar.com';
+
+		if ( 'mystery' == $default )
```
dkotter: Same as above

dinhtungdu:
@dkotter Currently we have a lot of coding style issues, I think it's better to address them in a separated PR after this is merged, does it work for you?

Log1x:
any update on this?

jeffpaul:
@Log1x we have a couple other plugins scheduled for releases ahead of Simple Local Avatars.  Getting this PR through review and into a release is on our todo list, just not in the immediate future.  If you have an urgent or important need for this specific change, then please reach out to me directly and we discuss... thanks!

joshuaabenazer:
So while using Simple Local Avatars on a project recently, I just needed the image url rather than the whole image tag. I assumed that since `get_avatar()` works out of the box, the other functions would also be intercepted similarly. The case in question here is the `get_avatar_url()` function.
We can use the `pre_get_avatar_data` filter in `get_avatar_data()` to short-circuit the function and updated the URL or maybe just let it go through the whole function and update the `args` accordingly in the `get_avatar_data` filter or maybe just `get_avatar_url` filter

I feel that if we update the URL at that stage, we may not be required to override the pluggable `get_avatar()` function. I haven't spent much time in this and I could be wrong, but I feel this is the right ticket to look into this and see if we can take over all the avatar related functions in core, just so that we are able to use them as is while the plugin does the rest behind the scenes.

helen (dismissed):

ammon-lockwood:
@dinhtungdu Thank you, I'm seeing the right URL as well if I use the `get_avatar_url` and the `get_avatar` functions. It appears the issue is between this plugin and the [wpgraphql](https://github.com/wp-graphql/wp-graphql) plugin which expects to have the `found_avatar` set. In the code in #40 when we find the local avatar and set it in the `$args` on line 81 we should also set the `found_avatar` to `true` right after on line 82 so that anything looking for that argument can key off it. In our case wpgraphql doesn't return anything from the avatar data unless `found_avatar` is `true`. WordPress sets it to true when the default Gravatar is set and so we should set this argument as well when we find the local avatar. I propose the `get_avatar_data` function look like this:

```php
        /**
	 * Filter avatar data early to add avatar url if needed. This filter hooks
	 * before Gravatar setup to prevent wasted requests.
	 *
	 * @since 2.2.0
	 *
	 * @param array $args        Arguments passed to get_avatar_data(), after processing.
	 * @param mixed $id_or_email The Gravatar to retrieve. Accepts a user ID, Gravatar MD5 hash,
	 *                           user email, WP_User object, WP_Post object, or WP_Comment object.
	 */
	public function get_avatar_data( $args, $id_or_email ) {
		$simple_local_avatar_url = $this->get_simple_local_avatar_url( $id_or_email, $args['size'] );
		if( $simple_local_avatar_url ) {
			$args['url'] = $simple_local_avatar_url;
			$args['found_avatar'] = true;
		}

		// Local only mode
		if( ! $simple_local_avatar_url && ! empty( $this->options['only'] ) ) {
			$args['url'] = $this->get_default_avatar_url( $args['size'] );
		}

		return $args;
	}
```

dinhtungdu:
@sumnercreations Can you please test the latest commit and see if your issue is resolved? https://github.com/10up/simple-local-avatars/tree/fix/39

jeffpaul:
After internal discussion and review, it was agreed this was ready for merge so I'm going to get this in and start looking towards what else remains in a 2.2.0 release.

Issue #42, opened 2020-03-20
Test against WP 5.4


jeffpaul:
**Is your enhancement related to a problem? Please describe.**
Once WordPress 5.4 is released, we'll want to test Simple Local Avatars to see if any incompatibility issues arise.

**Describe the solution you'd like**
- [ ] test Simple Local Avatars on WP 5.4
- [ ] open issues for any incompatibilities noted in testing
- [ ] resolve issues identified in testing
- [ ] bump "tested up to" version
- [ ] if code changes needed due to incompatibilities, ship a plugin release, otherwise use "Plugin asset/readme update" action to update "tested up to" version on .org repo

**Designs**
n/a

**Describe alternatives you've considered**
none

**Additional context**
none

jeffpaul:
Closed via #43.

PR #36, opened 2019-12-16
Plugin tested with WP 5.3

Waka867:
### Description of the Change

No code changes made. Plugin readme.txt updated to "Tested up to WP 5.3"

### Verification Process

Installed plugin
Plugin loads properly when activated
Profile Image uploads properly
Image is properly saved in uploads with full and five other image sizes
Image maturity rating saves and updates properly
No console errors
Avatar properly obeys max maturity rating set in Settings>Discussion
Avatar takes over profile image when image is appropriate according to maturity levels
Gravatar properly takes over if you have one and no local avatar is set
Only allow local avatars checkbox works properly
Checkbox to only allow users with file upload capabilities to upload local avatars is working properly
Avatar area resizes properly
Updated tested up to in readme.txt to 5.3

### Checklist:

- [ X] I have read the [**CONTRIBUTING**](https://github.com/10up/simple-local-avatars/blob/develop/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

PR #43, opened 2020-04-04
Bumping tested up to version.

ghost:
Bumping WP tested up to version to 5.4. I followed [the same verification process as here.](https://github.com/10up/simple-local-avatars/pull/36)

For #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 > Adds an avatar upload field to user profiles. Generates requested sizes on demand just like Gravatar!
 
-[![Support Level](https://img.shields.io/badge/support-active-green.svg)](#support-level) [![Release Version](https://img.shields.io/github/release/10up/simple-local-avatars.svg)](https://github.com/10up/simple-local-avatars/releases/latest) ![WordPress tested up to version](https://img.shields.io/badge/WordPress-v5.3%20tested-success.svg) [![GPLv2 License](https://img.shields.io/github/license/10up/simple-local-avatars.svg)](https://github.com/10up/simple-local-avatars/blob/develop/LICENSE.md)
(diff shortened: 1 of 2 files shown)


jeffpaul:
@helen we'll need to update this to merge into `master` to trigger the GitHub Action, yeah?

jeffpaul:
I went ahead and pushed the same change to `master` to trigger the readme/asset action, but nothing happened.  Turns out our actions in `master` are the old format and won't work.  Sigh, we'll have to wait for a release here to get things back in order on `master`.

helen:
@jeffpaul Should be okay to update the workflow file in master and it'll run when merging that :)

jeffpaul:
Thanks @helen that worked (see #44) though the odd thing is dotorg now shows tested-up-to 5.4.1 but our readme.txt shows 5.4. Not a major issue, just not sure how that discrepancy snuck in.

helen:
@jeffpaul .org only requires that we bump majors :)

jeffpaul:
@helen I'm aware of that, my head-scratcher was that our readme.txt states 5.4 while the plugin page on dotorg shows 5.4.1.  I'm not certain how dotorg inferred 5.4.1 from our stated 5.4?

helen:
@jeffpaul Right, we only bump majors, and .org assumes/displays the latest minor in that major branch

jeffpaul:
Huh, ok.  That's a bit odd and potentially concerning as theoretically there could be an conflict between a plugin and a WP minor, especially now that [minor releases can include enhancements](https://make.wordpress.org/core/handbook/about/release-cycle/releasing-minor-versions/), that would only be caught in specific testing against a minor versus against a major release.

PR #48, opened 2020-06-03
Change assign_new_user_avatar to public

### LINKED ISSUE #46: Ability to programmatically set avatar
Had a client contact me who is trying to set the avatar from this plugin, using a value from one of my plugins -- but it seems that the method to set the avatar is set to private:
https://github.com/10up/simple-local-avatars/blob/develop/simple-local-avatars.php#L329

And there doesn't really seem to be any other easy way to set this from a value outside of the plugin itself

tripflex:
fixes #46


jeffpaul:
Thanks for the PR @tripflex, I've milestoned this for our next release and we'll let you know if there are any questions or concerns once this gets through review... cheers!

dinhtungdu (approved):

PR #69, opened 2021-02-23
Fixed string as int in REST API

diodoe:
### Description of the Change

Bug: send media_id via REST api doesn't work.

### Benefits

You can now refer correctly to a media_id via REST api

### Possible Drawbacks

None

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "diodoe/simple-local-avatars",
+  "name": "10up/simple-local-avatars", 
   "description": "Adds an avatar upload field to user profiles. Generates requested sizes on demand just like Gravatar!",
   "type": "wordpress-plugin",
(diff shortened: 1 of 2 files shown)


dinhtungdu (changes requested):
Hi @diodoe, thanks for the PR. `is_numeric` returns `true` for float number, but we only want integer. Can you please update your PR to fix this issue?

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "10up/simple-local-avatars",
+  "name": "diodoe/simple-local-avatars",
```
dinhtungdu: Please change this back to `10up`.

dinhtungdu:
@diodoe Thanks for the update, I think we can solve this issue better, please review #71 and let me know.

PR #72, opened 2021-04-06
Add more robust multisite support and shared avatar setting

### LINKED ISSUE #26: Ensure avatars work correctly on a network install
#### Description
I would expect that the plugin would work on a network install.

#### Actual results
On a network install, after setting an avatar for my user, WordPress continued to use my gravatar instead.

dkotter:
### Description of the Change

This PR attempts to add more robust support for WordPress multisite. If this plugin is network activated, a new network settings section is added. This section contains the two current settings (`Local avatars only` and `Local upload permissions`) as well as a new setting, `Shared network avatars` (more on this later).

It also has an entirely new section to choose whether these settings should be the default for newly created sites or if they should be enforced across all sites. If the `default` option is chosen, individual sites on the network can change these settings on their respective sites. If `enforce` is chosen, all sites will use these settings.

For the new `Shared network avatars` setting, the default for this is to be turned on. This means if a user sets an avatar on a single site in the network, that avatar will be used on all sites in the network, since users are shared across the entire network. If this setting is turned off, users can have different avatars for each individual site.

### Screenshots
Screen Shot 2021-04-06 at 8 20 09 AM

### Alternate Designs

None

### Benefits

More robust support for the case of having the plugin network activated. In that scenario, adds the ability to either use shared avatars across all sites or use unique avatars on a site by site basis.

### Possible Drawbacks

There shouldn't be any, as this all should be backwards compatible with how things currently work.

### Verification Process

1. Test that everything works as before on a single site install
2. Also test that everything works as before on a network install without having the plugin network activated
3. Then on a network install, network activate the plugin
4. Verify new settings show up on the Network settings page 
5. Test out all the various combinations of settings (enforced, non-enforced, etc) and ensure everything works as expected
6. In particular, test out the new shared avatars feature and make sure that works both when turned on and off

### Checklist:

### Applicable Issues

#26


dinhtungdu:
@dkotter Testing this one and I found an issue updating avatar, here are steps to reproduce:

1. SLA is network-enabled.
2. Mode is set to enforce, Shared network avatars is enabled.
3. Go to site A (not network admin), set the avatar for current user. - Working.
4. Go to site B, log in as the same user, try changing the avatar - Not working, the new avatar is not set and the old one is removed across network.

holle75:
Hi there, sorry to barge in, but today i realized this problem. Unfortunately at the moment it seems impossible to manually set the same avatar on different sites at all. Whenever i change the avatar on one site, gravatars are getting used on the other site. Vice Verca.
Meaning the same NETWORKuser will never have the same manual picked avatar on all Multisites. This kind of makes using this plugin impossible on a multisite setup.

It would be amazing, if you guys could fix this.

jeffpaul:
@holle75 are you able to test this PR and note whether it works as you would expect or comment on what's still not ideal for you?

holle75:
Dear @jeffpaul, i would. But being a "github-newbie" and no skilled  programmer i don´t have the proper infrastructure to implement 8 commits. One i could manage manually, Modifying 8 files by hand might create chaos.

jeffpaul:
@dkotter any thoughts on @dinhtungdu's feedback in https://github.com/10up/simple-local-avatars/pull/72#issuecomment-815745220?

faisal-alvi:
@jeffpaul the issue mentioned by @dinhtungdu is fixed.

includes/class-simple-local-avatars.php
```diff
@@ -82,12 +171,12 @@ public function plugin_filter_action_links( $links ) {
 	/**
 	 * Retrieve the local avatar for a user who provided a user ID, email address or post/comment object.
 	 *
-	 * @param string            $avatar      Avatar return by original function
+	 * @param string $avatar Avatar return by original function
```
cadic: Type hints alignment is broken

cadic (changes requested):
Code looks good, except one small issue with composer.lock file and lot of code style errors/warnings, mostly not related to the current change.

composer.json
```diff
@@ -22,7 +22,6 @@
     "php": ">=5.3"
   },
   "require-dev": {
-    "10up/wpacceptance": "dev-master",
```
cadic, line 25: `composer.lock` still contain lot of references to wpacceptance and deps. Consider running `composer update` locally and commit the updated lock file here.

After `composer update && composer install` all tests will fail because PHPUnit 9.5 will be installed by default. Consider locking it with 7.5 in composer.json OR upgrading all tests to be compatible with PHPUnit 9.5.

faisal-alvi: thanks for the suggestion, implemented!

includes/class-simple-local-avatars.php
```diff
@@ -1,26 +1,47 @@
...
 			'PG' => __( 'PG &#8212; Possibly offensive, usually for audiences 13 and above', 'simple-local-avatars' ),
 			'R'  => __( 'R &#8212; Intended for adult audiences above 17', 'simple-local-avatars' ),
 			'X'  => __( 'X &#8212; Even more mature than above', 'simple-local-avatars' ),
 		);
 
-		$this->add_hooks();
+		if (
+			! $this->is_avatar_shared() // Are we sharing avatars?
+			&& (
+				( // And either an ajax request not in the network admin
```
cadic:
```suggestion
				( // And either an ajax request not in the network admin.
```

includes/class-simple-local-avatars.php
```diff
@@ -1,26 +1,47 @@
...
+					defined( 'DOING_AJAX' ) && DOING_AJAX
+					&& ! preg_match( '#^' . network_admin_url() . '#i', $_SERVER['HTTP_REFERER'] )
+				)
+				||
+				( // Or normal request not in the network admin
```
cadic:
```suggestion
				( // Or normal request not in the network admin.
```

includes/class-simple-local-avatars.php
```diff
@@ -55,6 +77,73 @@ public function add_hooks() {
...
+
+	/**
+	 * Determine if plugin is network activated.
+	 *
+	 * @param string $plugin The plugin slug to check.
+	 *
+	 * @return boolean
+	 */
+	public static function is_network( $plugin ) {
+		$plugins = get_site_option( 'active_sitewide_plugins', [] );
```
cadic:
```suggestion
		$plugins = get_site_option( 'active_sitewide_plugins', array() );
```

includes/class-simple-local-avatars.php
```diff
@@ -55,6 +77,73 @@ public function add_hooks() {
...
+	}
+
+	/**
+	 * Determine if avatars should be shared
+	 *
+	 * @return boolean
+	 */
+	public function is_avatar_shared() {
+		if (
+			is_multisite() // Are we on multisite
```
cadic:
```suggestion
			is_multisite() // Are we on multisite.
```

includes/class-simple-local-avatars.php
```diff
@@ -55,6 +77,73 @@ public function add_hooks() {
...
+			&& ! isset( $this->options['shared'] ) // And our shared option doesn't exist
```
cadic:
```suggestion
			&& ! isset( $this->options['shared'] ) // And our shared option doesn't exist.
```

includes/class-simple-local-avatars.php
```diff
@@ -55,6 +77,73 @@ public function add_hooks() {
...
+			|| (
+				isset( $this->options['shared'] ) // Or our shared option is set
```
cadic:
```suggestion
				isset( $this->options['shared'] ) // Or our shared option is set.
```

includes/class-simple-local-avatars.php
```diff
@@ -170,8 +259,20 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
 
 		// handle "real" media
 		if ( ! empty( $local_avatars['media_id'] ) ) {
+			// If using shared avatars, make sure we validate the URL on the main site
```
cadic:
```suggestion
			// If using shared avatars, make sure we validate the URL on the main site.
```

includes/class-simple-local-avatars.php
```diff
@@ -282,10 +384,35 @@ public function admin_init() {
...
+					'key'     => 'shared',
+					'desc'    => __( 'Uploaded avatars will be shared across the entire network, instead of being unique per site', 'simple-local-avatars' ),
+					'default' => 1,
+				)
+			);
+		}
+
+		add_action( 'load-options-discussion.php', array( $this, 'load_discussion_page' ) );
+
+		// This is for network site settings
```
cadic:
```suggestion
		// This is for network site settings.
```

includes/class-simple-local-avatars.php
```diff
@@ -306,6 +433,125 @@ public function admin_init() {
...
+	public function load_discussion_page() {
+		add_action( 'admin_print_styles', array( $this, 'admin_print_styles' ) );
+		add_filter( 'admin_body_class', array( $this, 'admin_body_class' ) );
+	}
+
+	/**
+	 * Load needed hooks to handle network settings
+	 */
+	public function load_network_settings() {
+		$this->options = (array) get_site_option( 'simple_local_avatars', [] );
```
cadic:
```suggestion
		$this->options = (array) get_site_option( 'simple_local_avatars', array() );
```

includes/class-simple-local-avatars.php
```diff
@@ -306,6 +433,125 @@ public function admin_init() {
...
+            </tr>
+        </table>
+
+		<?php
+	}
+
+	/**
+	 * Handle saving the network settings
+	 */
+	public static function save_network_settings() {
```
cadic, line 529: This function triggers some `WordPress.Security` warnings, not sure if we want to resolve every one or just add ignore rules to all of them.

includes/class-simple-local-avatars.php
```diff
@@ -395,6 +647,19 @@ public function avatar_settings_field( $args ) {
 			echo '<button id="clear_cache_btn" class="button delete" name="clear_cache_btn" >' . esc_html__( 'Clear cache', 'simple-local-avatars' ) . '</button><br/>';
 			echo '<span id="clear_cache_message" style="font-style:italic;font-size:14px;line-height:2;"></span>';
 		}
+
+		// Output warning if needed
```
cadic:
```suggestion
		// Output warning if needed.
```

includes/class-simple-local-avatars.php
```diff
@@ -395,6 +647,19 @@ public function avatar_settings_field( $args ) {
...
+		if (
+			SLA_IS_NETWORK // If network activated
```
cadic:
```suggestion
			SLA_IS_NETWORK // If network activated.
```

includes/class-simple-local-avatars.php
```diff
@@ -395,6 +647,19 @@ public function avatar_settings_field( $args ) {
...
+			&& $this->is_enforced() // And in enforce mode
```
cadic:
```suggestion
			&& $this->is_enforced() // And in enforce mode.
```

includes/class-simple-local-avatars.php
```diff
@@ -395,6 +647,19 @@ public function avatar_settings_field( $args ) {
...
+			&& 'shared' === $args['key'] // And we are displaying the last setting
```
cadic:
```suggestion
			&& 'shared' === $args['key'] // And we are displaying the last setting.
```

includes/class-simple-local-avatars.php
```diff
@@ -515,7 +780,7 @@ public function user_edit_form_tag() {
 	 */
 	public function assign_new_user_avatar( $url_or_media_id, $user_id ) {
 		// delete the old avatar
-		$this->avatar_delete( $user_id );    // delete old images if successful
+		$this->avatar_delete( $user_id ); // delete old images if successful
```
cadic:
```suggestion
		$this->avatar_delete( $user_id ); // delete old images if successful.
```

includes/class-simple-local-avatars.php
```diff
@@ -525,9 +790,10 @@ public function assign_new_user_avatar( $url_or_media_id, $user_id ) {
 			$url_or_media_id        = wp_get_attachment_url( $url_or_media_id );
 		}
 
-		$meta_value['full'] = $url_or_media_id;
+		$meta_value['full']    = $url_or_media_id;
+		$meta_value['blog_id'] = get_current_blog_id();
 
-		update_user_meta( $user_id, 'simple_local_avatar', $meta_value );    // save user information (overwriting old)
+		update_user_meta( $user_id, $this->user_key, $meta_value ); // save user information (overwriting old)
```
cadic:
```suggestion
		update_user_meta( $user_id, $this->user_key, $meta_value ); // save user information (overwriting old).
```

includes/class-simple-local-avatars.php
```diff
@@ -588,12 +856,12 @@ public function edit_user_profile_update( $user_id ) {
...
 
 		// Handle ratings
-		if ( isset( $avatar_id ) || $avatar = get_user_meta( $user_id, 'simple_local_avatar', true ) ) {
+		if ( isset( $avatar_id ) || $avatar = get_user_meta( $user_id, $this->user_key, true ) ) {
 			if ( empty( $_POST['simple_local_avatar_rating'] ) || ! array_key_exists( $_POST['simple_local_avatar_rating'], $this->avatar_ratings ) ) {
 				$_POST['simple_local_avatar_rating'] = key( $this->avatar_ratings );
 			}
 
-			update_user_meta( $user_id, 'simple_local_avatar_rating', $_POST['simple_local_avatar_rating'] );
+			update_user_meta( $user_id, $this->rating_key, $_POST['simple_local_avatar_rating'] );
```
cadic, line 864: Triggers `WordPress.Security` warnings

includes/class-simple-local-avatars.php
```diff
@@ -761,6 +1030,77 @@ public function set_avatar_rest( $input, $user ) {
...
+	/**
+	 * Short-circuit filter the `simple_local_avatars` option to match network if necessary
+	 *
+	 * @param bool $value Value of `simple_local_avatars` option, typically false.
+	 *
+	 * @return array
+	 */
+	public function pre_option_simple_local_avatars( $value ) {
+		if ( SLA_IS_NETWORK && 'enforce' === $this->get_network_mode() ) {
+			$value = get_site_option( 'simple_local_avatars', [] );
```
cadic:
```suggestion
			$value = get_site_option( 'simple_local_avatars', array() );
```

cadic (dismissed):
Thank you @faisal-alvi, the code looks good.

I left other code style suggestions regarding current change. But seems they does not display in the editor by default because phpcs.xml configured incorrectly (missing `<file>` directive). This might worth a separate issue.

Approving the PR!

includes/class-simple-local-avatars.php
faisal-alvi: Do they worth sharing?

faisal-alvi (approved):

PR #85, opened 2021-10-19
80 wp user avatar

### LINKED ISSUE #80: Allow import from WP User Avatar / ProfilePress
As seen on [WP.org](https://wordpress.org/support/topic/move-from-wp-user-avatar/#post-14657633), folks are migrating to Simple Local Avatars from WP User Avatar / ProfilePress and looking for help migrating in their users avatars.  

Leverage [Gist](https://gist.github.com/philipjohn/822d3521a95481f6ad7e118a7106fbc7) to help script this migration (properly crediting the author of course).

**Designs**
Possibly include an “Import settings from…” option in settings to import settings from competing plugins starting with WP User Avatar.

**Additional context**

claytoncollie:
Provides admin-ajax and wp-cli methods for migrating avatars away from the WP User Avatar plugin.

Addresses #80 

### Verification Process

Visit /wp-admin/options-discussion.php to find the button at the bottom of the page to use the admin-ajax version.

Or run `simple-local-avatars migrate wp-user-avatar` to use the wp-cli version.

In both cases, we create a database option `simple_local_avatars_migrations` to store previously migrated avatars.

### Checklist:

### Changelog Entry

* Add wp-cli command to migrate avatars from WP User Avatars
* Add dashboard setting to migrate avatars from WP User Avatars


claytoncollie:
@jeffpaul I had to get this over to you to at least start a conversation or else it would drag on too much. I know there is a lot of code duplication between the ajax version and the cli version. The CLI warning, error, and success messages sprinkled throughout the method made it a bit more difficult to re-use for the ajax version. I'm happy to DRY it out but wanted to at least get your eyes on it before wasting much more time. Happy to make any and all revisions you see fit.

includes/class-simple-local-avatars.php
```diff
@@ -259,15 +272,16 @@ public function admin_init() {
 	 * @param string $hook_suffix Page hook
 	 */
 	public function admin_enqueue_scripts( $hook_suffix ) {
-		if ( 'profile.php' !== $hook_suffix && 'user-edit.php' !== $hook_suffix ) {
+		if ( 'profile.php' !== $hook_suffix && 'user-edit.php' !== $hook_suffix && 'options-discussion.php' !== $hook_suffix ) {
```
helen: I think at this length of conditional checking we should switch to an `in_array()` and possibly make the array of screens filterable.

includes/class-simple-local-avatars.php
```diff
@@ -259,15 +272,16 @@ public function admin_init() {
...
 			return;
 		}
 
 		if ( current_user_can( 'upload_files' ) ) {
 			wp_enqueue_media();
 		}
 
-		$user_id = ( 'profile.php' === $hook_suffix ) ? get_current_user_id() : (int) $_GET['user_id'];
+		$param_user_id = isset( $_GET['user_id'] ) ? $_GET['user_id'] : 0;
```
helen: I don't think this will pass VIP review if this plugin is in use there, need to do something like an `absint()` or some other `filter_var` type thing on the `$_GET` var.

includes/class-simple-local-avatars.php
```diff
@@ -324,6 +339,16 @@ public function avatar_settings_field( $args ) {
 		';
 	}
 
+	/**
+	 * Settings field for migrating avatars away from WP User Avatars
+	 */
+	public function migrate_from_wp_user_avatar_settings_field() {
+		printf(
+			'<button type="button" name="simple-local-avatar-migrate-from-wp-user-avatars" id="simple-local-avatar-migrate-from-wp-user-avatars" class="button button-secondary">%s</button><span class="spinner"></span>',
```
helen: Need to switch the plural to `simple-local-avatars` and singular `wp-user-avatar` :) `wp-user-avatars` is another distinct plugin, fun times.

includes/class-simple-local-avatars.php
```diff
@@ -678,4 +703,152 @@ public function get_avatar_rest( $user ) {
...
+	 * This function creates a new option in the wp_options table to store the processed user IDs
+	 * so that we can run this command multiple times without processing the same user over and over again.
+	 *
+	 * Credit to Philip John for the Gist
+	 *
+	 * @see https://gist.github.com/philipjohn/822d3521a95481f6ad7e118a7106fbc7
+	 */
+	public function ajax_migrate_from_wp_user_avatar() {
+		// Check required information.
+		if ( empty( $_POST['_wpnonce'] ) || ! wp_verify_nonce( $_POST['_wpnonce'], 'migrate_from_wp_user_avatar_nonce' ) ) {
```
helen: I would suggest using a POST var name that is more specific to the plugin, just in case of collisions.

simple-local-avatars.dev.js
```diff
@@ -77,6 +77,16 @@ jQuery(document).ready(function($){
 			avatar_preview.attr( 'src', current_avatar );
 		}
 	} );
+
+	$( document.getElementById('simple-local-avatar-migrate-from-wp-user-avatars') ).on( 'click', function(event) {
+		event.preventDefault();
+        $('.spinner').addClass('is-active');
+		jQuery.post( ajaxurl, { action: 'migrate_from_wp_user_avatar', _wpnonce: i10n_SimpleLocalAvatars.wpUserAvatarNonce }, function(data) {
```
helen: Same note here about the `_wpnonce` var name.

simple-local-avatars.dev.js
```diff
@@ -77,6 +77,16 @@ jQuery(document).ready(function($){
 			avatar_preview.attr( 'src', current_avatar );
 		}
 	} );
+
+	$( document.getElementById('simple-local-avatar-migrate-from-wp-user-avatars') ).on( 'click', function(event) {
+		event.preventDefault();
+        $('.spinner').addClass('is-active');
```
helen: Need to use a more specific selector in case of multiple spinners on page.

simple-local-avatars.dev.js
```diff
@@ -77,6 +77,16 @@ jQuery(document).ready(function($){
 			avatar_preview.attr( 'src', current_avatar );
 		}
 	} );
+
+	$( document.getElementById('simple-local-avatar-migrate-from-wp-user-avatars') ).on( 'click', function(event) {
```
helen: Same thing about the plural/singular swap :)

simple-local-avatars.dev.js
```diff
@@ -77,6 +77,16 @@ jQuery(document).ready(function($){
...
+		event.preventDefault();
+        $('.spinner').addClass('is-active');
+		jQuery.post( ajaxurl, { action: 'migrate_from_wp_user_avatar', _wpnonce: i10n_SimpleLocalAvatars.wpUserAvatarNonce }, function(data) {
+			if ( data != '' ) {
+				$('.spinner').removeClass('is-active');
```
helen: Match selector to whatever you change above.

helen (changes requested):
Overall looks reasonable code-wise, I did not functionally test it for the moment. Left inline notes about smaller things, as for the DRY part I do think it's worth doing for consistency and sanity. You should be able to split the code paths within one method with the `WP_CLI` constant, and the AJAX version really should return a response status and message to show user feedback on the screen, e.g. "successfully migrated X avatars" or a descriptive error message.

claytoncollie:
@helen Thanks for the feedback. I refactored the migration method to return an integer based on how many avatars were processed. That method is called in the ajax or the wp-cli methods then the value is used in the messaging. I still have some work to do with javascript error messaging and spinner for the dashboard version.

simple-local-avatars.dev.js
claytoncollie: Removed

simple-local-avatars.dev.js
claytoncollie: Removed

claytoncollie:
@helen I added localized success and error messages to the admin version so the user can see how many avatars were processed. I am very close to finishing my work depending on your feedback. This could use another code review when you have a minute.

includes/class-simple-local-avatars.php
```diff
@@ -251,6 +257,13 @@ public function admin_init() {
 				'desc' => __( 'Only allow users with file upload capabilities to upload local avatars (Authors and above)', 'simple-local-avatars' ),
 			)
 		);
+		add_settings_field(
+			'simple-local-avatars-migration',
+			__( 'Migrations', 'simple-local-avatars' ),
```
helen: I'm not looking at this in the context of the full screen at the moment but I wonder if this might need to be a more descriptive section title like "Migrate Other Local Avatars" or something like that.

includes/class-simple-local-avatars.php
```diff
@@ -678,4 +717,159 @@ public function get_avatar_rest( $user ) {
...
+		if ( 0 === absint( $count ) ) {
+			WP_CLI::error( esc_html__( 'No avatars were migrated from WP User Avatar.', 'simple-local-avatars' ) );
+		}
+
+		WP_CLI::success(
+			sprintf(
+				'%s %s %s',
+				esc_html__( 'Successfully migrated', 'simple-local-avatars' ),
+				esc_html( $count ),
+				esc_html__( 'avatars from WP User Avatar.', 'simple-local-avatars' )
```
helen: Strings for translations should not be fragments - not all languages construct sentences the same way as English, translators will not have full context in GlotPress, and even in English you'll run into a plural problem if you migrate 1 avatar. You'll want to use `_n()` instead.

simple-local-avatars.dev.js
```diff
@@ -77,6 +77,33 @@ jQuery(document).ready(function($){
 			avatar_preview.attr( 'src', current_avatar );
 		}
 	} );
+
+	$( document.getElementById('simple-local-avatars-migrate-from-wp-user-avatar') ).on( 'click', function(event) {
+		event.preventDefault();
+		jQuery.post( ajaxurl, { action: 'migrate_from_wp_user_avatar', migrateFromWpUserAvatarNonce: i10n_SimpleLocalAvatars.migrateFromWpUserAvatarNonce } )
```
helen: Does this need to be `jQuery` here or can it just use `$`? Real question, I haven't traced through to see if I missed something in the scope.

tests/wpa/BasicTest.php
```diff
@@ -59,6 +59,10 @@ public function testAdminSettingShows() {
 
 		$I->seeText( 'Local Avatars Only' );
 
+		$I->seeText( 'Local Upload Permissions' );
+
+		$I->seeText( 'Migrations' );
```
helen: Will want to match this to the other header if you change it.

helen (commented):
Looking good overall - left a few comments about smaller things, mainly around strings.

includes/class-simple-local-avatars.php
claytoncollie-dmv: Done

tests/wpa/BasicTest.php
claytoncollie-dmv: Done

simple-local-avatars.dev.js
claytoncollie-dmv: This matches how another ajax request is being made on line 38. Most likely all of the instances of `jQuery` can be `$` since that parameter is being passed in on line 12. Leaving as is for now to match the current pattern.

includes/class-simple-local-avatars.php
claytoncollie-dmv: I change the wording to a single string for the CLI version as well as the ajax version. Makes things easier on everyone.

claytoncollie-dmv:
@jeffpaul This needs another round of review when you are ready.

faisal-alvi:
Hi @claytoncollie 

While reviewing the PR, I am facing a fatal error.

_Using Ajax:_

85-fatal-error-ajax


_Using WP-CLI:_

85-fatal-error-wp-cli

Can you please look into this? (I'm using a single site setup)

includes/class-simple-local-avatars.php
```diff
@@ -277,11 +300,15 @@ public function admin_enqueue_scripts( $hook_suffix ) {
...
-				'mediaNonce'       => wp_create_nonce( 'assign_simple_local_avatar_nonce' ),
+				'user_id'                             => $user_id,
+				'insertMediaTitle'                    => __( 'Choose an Avatar', 'simple-local-avatars' ),
+				'insertIntoPost'                      => __( 'Set as avatar', 'simple-local-avatars' ),
+				'deleteNonce'                         => $this->remove_nonce,
+				'mediaNonce'                          => wp_create_nonce( 'assign_simple_local_avatar_nonce' ),
+				'migrateFromWpUserAvatarNonce'        => wp_create_nonce( 'migrate_from_wp_user_avatar_nonce' ),
+				'migrateFromWpUserAvatarSuccess'      => __( 'Number of avatars successfully migrated from WP User Avatar', 'simple-local-avatars' ),
+				'migrateFromWpUserAvatarFailure'      => __( 'No avatars were migrated from WP User Avatar.', 'simple-local-avatars' ),
+				'migrateFromWpUserAvatarProgress'     => __( 'Migration in progress.', 'simple-local-avatars' ),
```
faisal-alvi: We can remove extra spaces before array double arrow(s).

includes/class-simple-local-avatars.php
```diff
@@ -678,4 +716,158 @@ public function get_avatar_rest( $user ) {
...
+	 * @return void
+	 */
+	public function ajax_migrate_from_wp_user_avatar() {
+		// Bail early if nonce is not available.
+		if ( empty( $_POST['migrateFromWpUserAvatarNonce'] ) ) {
+			die;
+		}
+
+		// Bail early if nonce is invalid.
+		if ( ! wp_verify_nonce( $_POST['migrateFromWpUserAvatarNonce'], 'migrate_from_wp_user_avatar_nonce' ) ) {
```
faisal-alvi: A non-sanitized input variable: `$_POST['migrateFromWpUserAvatarNonce']`

includes/class-simple-local-avatars.php
```diff
@@ -678,4 +716,158 @@ public function get_avatar_rest( $user ) {
...
+		foreach ( $sites as $site ) {
+			// Get the blog ID to use in the meta key and user query.
+			$blog_id = isset( $site->blog_id ) ? $site->blog_id : 1;
+
+			// Get the name of the meta key for WP User Avatar.
+			$meta_key = $wpdb->get_blog_prefix( $blog_id ) . 'user_avatar';
+
+			// Get processed users from database.
+			$migrations      = get_option( 'simple_local_avatars_migrations', array() );
+			$processed_users = isset( $migrations['wp_user_avatar'] ) ? $migrations['wp_user_avatar'] : array();
```
faisal-alvi, line 902: Can be replaced with `??` version.

includes/class-simple-local-avatars.php
```diff
@@ -678,4 +716,158 @@ public function get_avatar_rest( $user ) {
...
+		$sites = get_sites();
+
+		// Bail early if we don't find sites.
+		if ( empty( $sites ) ) {
+			return $count;
+		}
+
+		foreach ( $sites as $site ) {
+			// Get the blog ID to use in the meta key and user query.
+			$blog_id = isset( $site->blog_id ) ? $site->blog_id : 1;
```
faisal-alvi, line 895: Can be replaced with `??` version.

includes/class-simple-local-avatars.php
```diff
@@ -678,4 +716,158 @@ public function get_avatar_rest( $user ) {
...
+	 *
+	 *     $ wp simple-local-avatars migrate wp-user-avatar
+	 *     Success: Number of avatars successfully migrated from WP User Avatar: 5
+	 *
+	 * @param array|null $args       The arguments.
+	 * @param array|null $assoc_args The associative arguments.
+	 *
+	 * @return void
+	 */
+	public function wp_cli_migrate_from_wp_user_avatar( $args = null, $assoc_args = null ) {
```
faisal-alvi: Is it required to set default values?

faisal-alvi (changes requested):

includes/class-simple-local-avatars.php
claytoncollie-dmv: The [null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php) is only available in PHP 7.0 and above. The [minimum supported version of PHP for this plugin is 5.3](https://github.com/10up/simple-local-avatars/blob/develop/simple-local-avatars.php#L8).

includes/class-simple-local-avatars.php
claytoncollie-dmv: The [null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php) is only available in PHP 7.0 and above. The [minimum supported version of PHP for this plugin is 5.3](https://github.com/10up/simple-local-avatars/blob/develop/simple-local-avatars.php#L8).

includes/class-simple-local-avatars.php
claytoncollie: Resolved by removing the default values.

includes/class-simple-local-avatars.php
claytoncollie: Won't fix.

includes/class-simple-local-avatars.php
claytoncollie: Won't fix

includes/class-simple-local-avatars.php
claytoncollie: Used `sanitize_text_field` to fix this.

includes/class-simple-local-avatars.php
claytoncollie: Done

faisal-alvi (dismissed):

claytoncollie:
@faisal-alvi Please unapprove this PR. I have not fixed the major issue of it not working on single-site installs. `get_sites` is only available for multisite installs so I need to refactor a bit.

claytoncollie:
@faisal-alvi This is ready for review. I added support for single-site installs.

### Testing Steps

1. Install [WP user Avatar](https://wordpress.org/plugins/wp-user-avatar/) 
2. Create a new user and give them an avatar image in the WP User Avatars field. 
3. Run the WP CLI command or use the ajax button in the admin to migrate it. 
```wp simple-local-avatars migrate wp-user-avatar --yes```
or
```wp simple-local-avatars migrate wp-user-avatar```
4. Check the database table `*_usermeta` for `simple_local_avatar` to see if the image migrated properly. 
5. Check the database table `*_options` for `simple_local_avatars_migrations` to see if the user ID was saved.

Repeat steps for single site install and multisite install.

File: includes/class-simple-local-avatars.php [hunk 18]
```diff
@@ -678,4 +716,168 @@ public function get_avatar_rest( $user ) {
...
+		if ( is_multisite() ) {
+			$sites = get_sites();
+		} else {
+			$site          = new stdClass();
+			$site->blog_id = 1;
+			$sites         = array( $site );
+		}
+
+		// Bail early if we don't find sites.
+		if ( empty( $sites ) ) {
```
faisal-alvi, line 889: Are there chances of `$sites` being `empty`?

faisal-alvi (approved):
LGTM! Just a question asked, good to be merged.

includes/class-simple-local-avatars.php
claytoncollie: @faisal-alvi I don't believe so but I like to add those as a safe guard before any `foreach` or `while` loops just in case something does go wrong.

claytoncollie:
@faisal-alvi I am not sure what the release process is for this plugin BUT all of the JS from this PR lives in the `.dev` file so that probably needs to be compiled into the production `simple-loca-avatars.js` file. I also have not updated the readme.

faisal-alvi:
@claytoncollie I have re-checked the functionality. When trying to follow [these](https://github.com/10up/simple-local-avatars/pull/85#issuecomment-996835970) steps, on step no 2, I can not select any image for the avatar. It works for the `develop` branch. Please have a look.

https://user-images.githubusercontent.com/25176325/146909363-47ae00e9-839c-449e-ac2f-0e6688a93d9c.mp4

faisal-alvi (changes requested):
Re-requesting changes to fix the avatar selection issue.

faisal-alvi:
@claytoncollie regarding your JS question, we use `uglyjs` to minify it only, please use it and can you please also add the `npm` command to do it?

claytoncollie:
@faisal-alvi I reverted how we are grabbing the `user_id` on the `edit-user.php` screen. That was breaking the javascript on that screen.

https://github.com/10up/simple-local-avatars/blob/9499974b51ff5e3363e29786fb37871618e1b025/includes/class-simple-local-avatars.php#L284

`(int) $_GET('user_id')` works fine. I changed that to `get_query_var('user_id')` which is what broke it. This is not code that I originally wrote and was instructed to alter to not have `$_GET` in the code. I think we should leave this as is. It is not related to this PR.

As for the minified JS, I do not feel comfortable adding that to this PR. There is no package.json file in this plugin. Can you all take care of minification in the release branch in case there are other changes from other PRs?

faisal-alvi (changes requested):
Thanks for the changes @claytoncollie.

1. Let's leave the `(int) $_GET('user_id')` which is not related to the current PR.
2. I have added `package.json` and also added `uglifyjs` (in #94) so we just have to run `npm run uglifyjs` and it will minify the dev JS file and update the content of the `simple-local-avatars.js` file (which is a minfied version of the dev file).

Leaving this PR until #94 is merged, then you can run the command and we are good to merge.

faisal-alvi:
@jeffpaul we should add minifying the JS file as a checklist item in our release process. After #94 merge, we can use `npm run uglifyjs` command to minify JS at the time of release.

jeffpaul:
@faisal-alvi seems like maybe we could add that to https://github.com/10up/simple-local-avatars/blob/develop/.github/workflows/push-deploy.yml and potentially remove the minified JS file from the repo as that'll get generated as part of the WP.org deploy action?

faisal-alvi:
@jeffpaul that's great! done => https://github.com/10up/simple-local-avatars/pull/94/commits/2064ba0df37f6cc62841178299d740c5f78fb24a

faisal-alvi:
@claytoncollie I have merged `develop` to the `80-wp-user-avatar` branch to resolve the conflicts. The PR is working fine. However, I have noticed a couple of things while testing.

1. The migration works only once. After the first migration, if a user re-uses the image from the "WP User Avatar" field and re-try to migrate, it does not work.
2. When a user is using an avatar from the "WP User Avatar" plugin, then if he tries to upload an image in the SLA avatar field, a new image does not reflect. When we deactivate the "WP User Avatar" plugin or remove the "WP User Avatar" image, the SLA avatar starts displaying. It seems like "WP User Avatar" has higher priority over SLA avatars.

@jeffpaul can you please let me know if these are intentional behaviors or should be considered bugs?

jeffpaul:
1. Not sure I understand this flow, could you please restate?
2. That seems like something that should be changed, if someone attempts to change the image via the standard SLA flow then that image should take priority over whatever was migrated in from WP User Avatar.

faisal-alvi:
**Point no 1:**
> The migration works only once. After the first migration, if a user re-uses the image from the "WP User Avatar" field and re-try to migrate, it does not work.

* For example, a site admin is using "WP User Avatar" and the site has 100 users with avatars.
* Admin wants to move to SLA.
* SLA installed & activated.
* Admin visits Settings > Discussion and migrates avatars to SLA.
image
* But admin still uses "WP User Avatar" for 10 new users and 5 existing users.
* After some time, the admin wants to re-migrate and visits Settings > Discussion and clicks migrate button.
* Expectation would be: 10 new + 5 existing user avatars should be migrated.
* Reality is: only 10 new user avatars migrated.
* SLA does not re-migrate 5 exiting users as they are already "marked as migrated" somewhere in DB.
* Is this behavior expected or a bug?

**Point no 2:**
> When a user is using an avatar from the "WP User Avatar" plugin, then if he tries to upload an image in the SLA avatar field, a new image does not reflect. When we deactivate the "WP User Avatar" plugin or remove the "WP User Avatar" image, the SLA avatar starts displaying. It seems like "WP User Avatar" has higher priority over SLA avatars.

* For example, both plugins (WP User Avatar & SLA) are installed and active on the site.
* Admin sets an avatar using the "WP User Avatar" field for a user.
image
* After clicking "update", the SLA field also displays the same avatar set in the "WP User Avatar" field.
* After some time, the admin sets a different avatar, but this time using the "SLA" avatar field.
* The different avatar set via SLA does not display. An old avatar from "WP User Avatar" still displays (seems due to a higher priority).
* If admin removes avatar from "WP User Avatar" field OR deactivates "WP User Avatar" plugin, the new SLA avatar starts displaying!
* Is this behavior expected or a bug?

claytoncollie:
@faisal-alvi Part 1 is intended functionality. That is the entire reason behind storing the IDs in the options table. Our thinking was that if the site has a large number of users to migrate, the admin-ajax function might timeout before completion. And in that case, if it times out halfway through we have a record and can start over at the end of the record instead of from the very beginning. This might have been spoken about in this thread or on a zoom with @helen 

Part 2 seems to be a "feature" of WP User Avatar. That does make sense to me that a plugin name starting with the letter `W` is loading after a plugin name starting with `S` thus firing all of its filter and actions after SLA. I think if people have both plugins installed, the intended outcome is that they migrate and then deactivate the WPUA plugin. Am I giving the average user too much credit? One thing this plugin does not do is delete or modify any WPUA data. That might be a solution but we felt it was not our responsibility.

faisal-alvi (approved):

faisal-alvi:
@claytoncollie thanks for the answers, @jeffpaul we are good to go 🚀

PR #89, opened 2021-11-17
Avatar rating text is not translated properly if a user has a custom language Set

### LINKED ISSUE #88: Avatar rating text is not translated properly if a user has a custom language set
**Describe the bug**
Reported [here](https://wordpress.org/support/topic/translation-of-ratings-doesnt-work-for-user-language/). The Avatar Rating text is translated properly if a site language is set but not if a user language is set.

Looking into it, I believe it's related to this: https://make.wordpress.org/core/2016/11/07/changed-loading-order-for-current-user-in-4-7/. Basically where we have those [strings set](https://github.com/10up/simple-local-avatars/blob/develop/includes/class-simple-local-avatars.php#L15) is loaded prior to the user object being initialized, so the user's set language is never available and thus won't be considered when translating.

There may be other approaches but it may be as simple as moving those text strings to the actual function that needs them, which should be late enough in the loading process to fix the problem. We do use the avatar ratings (G, PG, etc) in other places, so those need to stay but the actual text strings are only used in a single location.

**Steps to Reproduce**

1. Select a site language other than English (es_ES for instance). Ensure proper language packs get installed
2. Go to your profile page and note the Avatar Rating text is translated
3. Turn site language back to english
4. Set user language to a different language (es_ES again)
5. Note the Avatar Rating text is not translated

**Expected behavior**
All text should be translated if a different language is chosen and translations are available

ActuallyConnor:
### Description of the Change

Adding string translation method around each `$avatar_rating` when it is output in the `edit_user_profile()` method.

### Verification Process

- [x] Site language is set to English

Screen Shot 2021-11-17 at 9 29 43 AM

- [x] User language is set to French

Screen Shot 2021-11-17 at 9 29 37 AM

- [x] Rating strings are translated to French

Screen Shot 2021-11-17 at 9 29 28 AM

### Checklist:

### Applicable Issues

[https://github.com/10up/simple-local-avatars/issues/88](https://github.com/10up/simple-local-avatars/issues/88)

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -408,10 +408,10 @@ class Simple_Local_Avatars {
 
                     // These are the same as $this->avatar_ratings but in order for string translations to appear we have to output them like the following
                     // We also use the default text domain here
-                                        echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'G' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'G', false ) . "/>" . __( 'G &#8212; Suitable for all audiences' ) . "</label><br />";
-                                        echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'PG' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'PG', false ) . "/>" . __( 'PG &#8212; Possibly offensive, usually for audiences 13 and above' ) . "</label><br />";
-                                        echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'R' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'R', false ) . "/>" . __( 'R &#8212; Intended for adult audiences above 17' ) . "</label><br />";
-                                        echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'X' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'X', false ) . "/>" . __( 'X &#8212; Even more mature than above' ) . "</label><br />";
+                    echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'G' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'G', false ) . "/>" . __( 'G &#8212; Suitable for all audiences' ) . "</label><br />";
+                    echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'PG' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'PG', false ) . "/>" . __( 'PG &#8212; Possibly offensive, usually for audiences 13 and above' ) . "</label><br />";
+                    echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'R' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'R', false ) . "/>" . __( 'R &#8212; Intended for adult audiences above 17' ) . "</label><br />";
+                    echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'X' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'X', false ) . "/>" . __( 'X &#8212; Even more mature than above' ) . "</label><br />";
 					?>
 						<p class="description"><?php esc_html_e( 'If the local avatar is inappropriate for this site, Gravatar will be attempted.', 'simple-local-avatars' ); ?></p>
 					</fieldset></td>
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -407,7 +407,7 @@ class Simple_Local_Avatars {
 					}
 
 					foreach ( $this->avatar_ratings as $key => $rating ) :
-						echo "\n\t<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( $key ) . "' " . checked( $profileuser->simple_local_avatar_rating, $key, false ) . "/> $rating</label><br />";
+						echo "\n\t<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( $key ) . "' " . checked( $profileuser->simple_local_avatar_rating, $key, false ) . "/>" . __($rating) . "</label><br />";
 					endforeach;
 					?>
 						<p class="description"><?php esc_html_e( 'If the local avatar is inappropriate for this site, Gravatar will be attempted.', 'simple-local-avatars' ); ?></p>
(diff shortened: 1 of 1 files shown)


includes/class-simple-local-avatars.php
```diff
@@ -407,7 +407,7 @@ class="button item-delete submitdelete deletion"
 					}
 
 					foreach ( $this->avatar_ratings as $key => $rating ) :
-						echo "\n\t<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( $key ) . "' " . checked( $profileuser->simple_local_avatar_rating, $key, false ) . "/> $rating</label><br />";
+						echo "\n\t<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( $key ) . "' " . checked( $profileuser->simple_local_avatar_rating, $key, false ) . "/>" . __($rating) . "</label><br />";
```
faisal-alvi: @ActuallyConnor Thank you for raising a PR!

This looks good and fixes the issue but as per standards, the plain strings should always be directly passed as a string literal in the translation function, not as a variable. Read more [here](https://developer.wordpress.org/reference/functions/__/#comment-540).

> This seems to work, but it will interfere in automatic parsing of your plugin/theme’s files for translation.

Moreover, we are already translating strings.

https://github.com/10up/simple-local-avatars/blob/86ce042ae571583986d0c439675fc3f9be5be04a/includes/class-simple-local-avatars.php#L15-L20

But these are loaded prior to the user object being initialized as mentioned in #88. That needs to be fixed in a standard way.

faisal-alvi (changes requested):

includes/class-simple-local-avatars.php
ActuallyConnor: Understood, I can work on fixing that to work in a standard way

ActuallyConnor:
@faisal-alvi I implemented a new solution and left some comments for clarity. Let me know if you have any questions or concerns.

includes/class-simple-local-avatars.php
```diff
@@ -406,9 +406,12 @@ class="button item-delete submitdelete deletion"
...
+                                        echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'G' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'G', false ) . "/>" . __( 'G &#8212; Suitable for all audiences' ) . "</label><br />";
+                                        echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'PG' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'PG', false ) . "/>" . __( 'PG &#8212; Possibly offensive, usually for audiences 13 and above' ) . "</label><br />";
+                                        echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'R' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'R', false ) . "/>" . __( 'R &#8212; Intended for adult audiences above 17' ) . "</label><br />";
+                                        echo "<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( 'X' ) . "' " . checked( $profileuser->simple_local_avatar_rating, 'X', false ) . "/>" . __( 'X &#8212; Even more mature than above' ) . "</label><br />";
```
faisal-alvi: Moving it to a separate function and updating `$this->avatar_ratings`, and calling just before the `foreach` starts - would be a better improvement. Then no need to change anything in the `foreach` at all.

The function would just have the same array from `__constructor`, overwriting the existing `$this->avatar_ratings`.

https://github.com/10up/simple-local-avatars/blob/86ce042ae571583986d0c439675fc3f9be5be04a/includes/class-simple-local-avatars.php#L15-L20

Also, using `esc_html` over `$ratings` would be an extra improvement.

https://github.com/10up/simple-local-avatars/blob/86ce042ae571583986d0c439675fc3f9be5be04a/includes/class-simple-local-avatars.php#L410

faisal-alvi (changes requested):

faisal-alvi:
@ActuallyConnor Thank you for the improved solution. I've left some suggestions, please check.

ActuallyConnor:
@faisal-alvi I've made some changes based on your recommendations, let me know what you think

includes/class-simple-local-avatars.php
```diff
@@ -401,15 +401,19 @@ class="button item-delete submitdelete deletion"
...
-					}
+						<?php
+						$this->update_avatar_ratings();
 
-					foreach ( $this->avatar_ratings as $key => $rating ) :
-						echo "\n\t<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( $key ) . "' " . checked( $profileuser->simple_local_avatar_rating, $key, false ) . "/> $rating</label><br />";
-					endforeach;
-					?>
+						if ( empty( $profileuser->simple_local_avatar_rating ) || ! array_key_exists( $profileuser->simple_local_avatar_rating,
+								$this->avatar_ratings ) ) {
```
faisal-alvi: 407 | ERROR | [x] Opening parenthesis of a multi-line function call must be the last content on the line
408 | ERROR | [x] Closing parenthesis of a multi-line function call must be on a line by itself

includes/class-simple-local-avatars.php
```diff
@@ -401,15 +401,19 @@ class="button item-delete submitdelete deletion"
...
+							$profileuser->simple_local_avatar_rating = 'G';
+						}
+
+						foreach ( $this->avatar_ratings as $key => $rating ) :
+							echo "\n\t<label><input type='radio' name='simple_local_avatar_rating' value='" . esc_attr( $key ) . "' " . checked( $profileuser->simple_local_avatar_rating,
+									$key, false ) . "/>" . esc_html( $rating ) . "</label><br />";
```
faisal-alvi: 413 | ERROR | [x] Opening parenthesis of a multi-line function call must be the last content on the line
414 | ERROR | [x] Closing parenthesis of a multi-line function call must be on a line by itself

includes/class-simple-local-avatars.php
```diff
@@ -678,4 +682,18 @@ public function get_avatar_rest( $user ) {
...
+     * Default text-domain because the strings have already been translated
+     */
+    private function update_avatar_ratings() {
+        $this->avatar_ratings = array(
+            'G'  => __( 'G &#8212; Suitable for all audiences' ),
+            'PG' => __( 'PG &#8212; Possibly offensive, usually for audiences 13 and above' ),
+            'R'  => __( 'R &#8212; Intended for adult audiences above 17' ),
+            'X'  => __( 'X &#8212; Even more mature than above' ),
+        );
+    }
```
faisal-alvi: 684-698 | ERROR | [x] Tabs must be used to indent lines; spaces are not allowed

faisal-alvi (changes requested):
@ActuallyConnor the solution looks good, thank you. There are a couple of errors from `PHPCS`, please fix them. Then we will be good to go.

ActuallyConnor:
@faisal-alvi those PHPCS errors should be fixed now 😄

faisal-alvi (approved):

PR #90, opened 2021-11-19
Add option to clear cache of user meta to remove image sizes that do not exist

### LINKED ISSUE #31: avatars are being deleted after thumbnail regenarting
after regenerating all thumbnails in site and wiping unnecessary images (using Regenerate Thumbnails plugin), some avatars thumbnail where deleted.  so...
1. is there a way to regenerate them? the original files are still there, only the cropped ones are gone.
2. there must a be a way to prevent things like that from happening. maybe registering the avatars as image_size?

thrijith:
### Description of the Change

- Adds a `Clear Cache` button in  `Settings` -> `Discussion`, which updates the user meta to remove images that don't exist.

### Alternate Designs

N/A

### Benefits

- Fixes issue with broken avatar links after media regeneration which may delete the images.

### Possible Drawbacks

N/A

### Verification Process

- Set an avatar for your profile
- Install `Regenerate Thumbnails` plugin
- Use `Tools` -> `Regenerate Thumbnails`, select `Delete thumbnail files for old unregistered sizes in order to free up server space. This may result in broken images in your posts and pages.` checkbox and regenerate the thumbnails.
- This will make the avatar images broken, go to `Settings` -> `Discussion` and use the `Clear Cache` to fix the issue.

### Checklist:

- [x] All new and existing tests passed.

### Applicable Issues

Fixs https://github.com/10up/simple-local-avatars/issues/31

### Changelog Entry

- Add `Clear Cache` button to fix broken avatar images after thumbnail regeneration.

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -1,73 +1,58 @@
-/* global ajaxurl, slaAdmin */
-(function ($) {
-    $(function () {
-        function slaAdminAction() {
-            const $sla = {
-                clearCache: $('#clear_cache_btn'),
-            };
+/* global ajaxurl, slaAdmin, jQuery */
+( function ( $ ) {
+	// Cache the button.
+	const $clearCacheBtn = $( '#clear_cache_btn' );
 
-            // Spinner button.
-            const spinnerButton = '<span class="spinner is-active"></span>';
+	// Spinner button.
+	const spinnerButton = '<span class="spinner is-active"></span>';
 
-            // Bind events.
-            const bindEvents = function () {
-                $sla.clearCache.on('click', function (e) {
-                    e.preventDefault();
-                    $sla.clearCache.addClass( 'disabled' );
-                    var data = {
-                        action: 'sla_clear_user_cache',
-                        nonce: slaAdmin.nonce,
-                    }
-                    $sla.clearCache.append(spinnerButton);
-                    processStep(1, data, $sla.clearCache);
-                });
-            };
+	// Bind events.
+	$clearCacheBtn.on( 'click', function ( e ) {
+		e.preventDefault();
+		$clearCacheBtn.addClass( 'disabled' );
+		$clearCacheBtn.append( spinnerButton );
+		var data = {
+			action: 'sla_clear_user_cache',
+			nonce: slaAdmin.nonce,
+		};
+		processStep( 1, data );
+	} );
 
-            /**
-             * Process steps.
-             *
-             * @param {Number} step For batch wise updates.
-             * @param {Object} data Data to pass to ajax request.
-             * @param {HTMLElement} el Spinner button.
-             */
-            function processStep(step, data, el) {
-                data.step = step;
-                $.ajax({
-                    url    : ajaxurl,
-                    dataType: 'json',
-                    data   : data,
-                    method : 'POST',
-                    success: function (response) {
-                        if (response.success) {
-                            if (response.data.step === 'done') {
-                                el.find('span').remove();
-                                $sla.clearCache.removeClass( 'disabled' );
-                            } else {
-                                processStep(parseInt(response.data.step, 10), data, el);
-                            }
-                            return false;
-                        } else {
-                            alert(slaAdmin.error);
-                            el.find('span').remove();
-                            $sla.clearCache.removeClass( 'disabled' );
-                        }
-                    },
-                    error  : function () {
-                        alert(slaAdmin.error);
-                        el.find('span').remove();
-                        $sla.clearCache.removeClass( 'disabled' );
-                    }
-                });
-            }
+	function removeSpinner() {
+		$clearCacheBtn.find( 'span' ).remove();
+		$clearCacheBtn.removeClass( 'disabled' );
+	}
 
-            const initializePage = function () {
-                bindEvents();
-            };
-
-            initializePage();
-        }
-
-        // Initialize page.
-        slaAdminAction();
-    });
-})(jQuery); // eslint-disable-line no-undef
+	/**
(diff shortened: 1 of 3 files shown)


dinhtungdu (commented):
@thrijith In this PR, we're deleting cache for the current admin user, normal users can't delete the cache because they don't have access to the setting page. Besides, the main usage of this feature is deleting the cache after regenerating thumbnails, which only admins can perform.

I think the deleting cache function should delete the avatar cache for all users instead. We also need to care about the performance of large sites that have many users.

thrijith:
> @thrijith In this PR, we're deleting cache for the current admin user, normal users can't delete the cache because they don't have access to the setting page. Besides, the main usage of this feature is deleting the cache after regenerating thumbnails, which only admins can perform.
> 
> I think the deleting cache function should delete the avatar cache for all users instead. We also need to care about the performance of large sites that have many users.

@dinhtungdu I've made some changes, please check again when you get a chance, thanks!

dinhtungdu (commented):
@thrijith This is likely to have timeout issues IMO. For large sites with a lot of users, it would take time to get all users, check and update the meta for each user. We're doing everything in a page load, which is not reliable to me. I think Ajax can be the solution to this issue.

thrijith:
> @thrijith This is likely to have timeout issues IMO. For large sites with a lot of users, it would take time to get all users, check and update the meta for each user. We're doing everything in a page load, which is not reliable to me. I think Ajax can be the solution to this issue.

@dinhtungdu please check again when you get a chance, thanks!

includes/class-simple-local-avatars.php
```diff
@@ -694,4 +712,100 @@ private function update_avatar_ratings() {
...
+					'step' => $step + 1,
+				)
+			);
+		}
+
+		wp_send_json_success(
+			array(
+				'step' => 'done',
+			)
+		);
```
dinhtungdu: I think we should send more detailed error messages to clients instead of displaying a generic one. That will make debugging much easier when something goes wrong.

assets/js/admin.js
```diff
@@ -0,0 +1,73 @@
...
+                            }
+                            return false;
+                        } else {
+                            alert(slaAdmin.error);
+                            el.find('span').remove();
+                            $sla.clearCache.removeClass( 'disabled' );
+                        }
+                    },
+                    error  : function () {
+                        alert(slaAdmin.error);
```
dinhtungdu: Instead of `alert`, can we print the error message below the clear cache button? By doing so we can display more detailed errors. We can even add the status there too, for example:

image

includes/class-simple-local-avatars.php
```diff
@@ -694,4 +712,100 @@ private function update_avatar_ratings() {
...
+		if ( ! empty( $users ) ) {
+			foreach ( $users as $user ) {
+				$user_id       = $user->ID;
+				$rows[]        = $user_id;
+				$local_avatars = get_user_meta( $user_id, 'simple_local_avatar', true );
+				$this->clear_user_avatar_cache( $local_avatars, $user_id, $local_avatars['media_id'] ?? '' );
+			}
+		}
+
+		if ( ! empty( $rows ) ) {
```
dinhtungdu: Instead of creating `$row` to hold only user IDs, I think we can use the `$users` variable above to check for the end of the progress.

dinhtungdu (commented):
@thrijith thanks for the update! I left some comments on your PR, please take a look. I also format the admin.js file in #91, please check it as well. Feel free to merge it if it makes sense to you.

includes/class-simple-local-avatars.php
thrijith: @dinhtungdu I've updated this, please check again when you get a chance, thanks!

assets/js/admin.js
thrijith: Added the message span below the button, demo https://cln.sh/u5NhwT

includes/class-simple-local-avatars.php
thrijith: @dinhtungdu didn't get this, at what point is the error needed message to be added?

dinhtungdu: I reviewed the ajax callback, looks like there isn't an error to be thrown there.

includes/class-simple-local-avatars.php
```diff
@@ -316,12 +330,17 @@ public function avatar_settings_field( $args ) {
...
+		if ( 'clear_cache' !== $args['key'] ) {
+			echo '
 			<label for="simple-local-avatars-' . esc_attr( $args['key'] ) . '">
 				<input type="checkbox" name="simple_local_avatars[' . esc_attr( $args['key'] ) . ']" id="simple-local-avatars-' . esc_attr( $args['key'] ) . '" value="1" ' . checked( $this->options[ $args['key'] ], 1, false ) . ' />
 				' . esc_html( $args['desc'] ) . '
 			</label>
 		';
+		} else {
+			echo '<button id="clear_cache_btn" class="button delete" name="clear_cache_btn" >' . esc_html__( 'Clear cache', 'simple-local-avatars' ) . '</button><br/>';
+			echo '<span id="clear_cache_message" style="font-style:italic;font-size:15px;line-height:2;"></span>';
```
dinhtungdu: I think we should follow the font size of other field in the setting page.
```suggestion
			echo '<span id="clear_cache_message" style="font-style:italic;font-size:14px;line-height:2;"></span>';
```

dinhtungdu (dismissed):
LGTM! Left an optional comment but this is ready to be merged!

dinhtungdu (approved):

PR #94, opened 2022-01-04
Introducing Package file

faisal-alvi:
### Description of the Change

The package file is introduced.

### Benefits

Helpful in cases where a new `script` is required like https://github.com/10up/simple-local-avatars/pull/85#issuecomment-1000456744. Added `uglifyjs` script to overcome the mentioned issue.

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.


jeffpaul (changes requested):
@faisal-alvi you'll want to add those files to https://github.com/10up/simple-local-avatars/blob/develop/.gitattributes so that they don't end up in the released version.

claytoncollie:
@faisal-alvi A minor suggestion, but renaming `uglifyjs` to `build` would be a better name to build the assets.

faisal-alvi:
@claytoncollie Thanks for the suggestion, I had preserved the “build” for the actual building of the “whole project” whereas the `uglifyjs` is minimizing only one JS file. But as we do not have any other files to be built (for now), we can utilize the word “build”.

jeffpaul:
Odd that the previous commit passed checks but the latest failed.  Given that its WP Acceptance (and that we should be setting up Cypress e2e tests here anyway) I think I'm fine with this for now (ideally we get those e2e tests in place before the next release).

cadic (approved):

PR #96, opened 2022-01-06
Ability to set default avatar

### LINKED ISSUE #78: Ability to set default avatar
Noting the following received via [10up.com](https://10up.com/plugins/simple-local-avatars-wordpress/comment-page-10/#comment-1126022):
> One feature i would love too see added / while not taking away from the light weight nature is…
An option to set the Default image used if someone has not yet uploaded an image.
>
> Currently it defaults to options from Gravatar but it would be really nice if we could set a custom default image :)

We'll want to consider the following:
- [ ] is this sort of feature worthwhile in this otherwise simple, lightweight plugin?
- [ ] should this be a per-site option, enabled network-wide, or handled some other way?

**Designs**

**Additional context**

mehulkaklotar:
### Benefits

It will allow admins to use custom default image avatars for users instead of using gravatar defaults.

### Checklist:


faisal-alvi:
Please note I have merged `develop` in the `feature/default-avatar` to resolve conflicts.

mehulkaklotar:
Hi @faisal-alvi 
can you run a quick look on the changes if they can pass or not? I have tested the default avatar in local but I would like to be tested more.
Thanks.

assets/js/admin.js
```diff
@@ -3,6 +3,9 @@
 	// Cache the button.
 	const $clearCacheBtn = $( '#clear_cache_btn' );
 	const $clearCacheMessage = $( '#clear_cache_message' );
+	const $simpleLocalAvatarUpload = $('#simple-local-avatar-upload');
```
faisal-alvi: Can we use something more specific, for example `simpleLocalAvatarDefault` and `#simple-local-avatar-default`?

assets/js/admin.js
```diff
@@ -58,4 +61,41 @@
...
+			library : {
+				type : 'image',
+			}
+		}).open()
+			.on('select', function(e){
+				// This will return the selected image from the Media Uploader, the result is an object
+				var uploaded_image = image.state().get('selection').first();
+				uploaded_image = uploaded_image.toJSON();
+				var avatar_preview = uploaded_image?.sizes?.thumbnail?.url;
+				if ( typeof avatar_preview === 'undefined' ) {
```
faisal-alvi:
```suggestion
				if ( 'undefined' === typeof avatar_preview ) {
```

assets/js/admin.js
```diff
@@ -58,4 +61,41 @@
...
+			});
+	});
+
+	if ( $simpleLocalAvatarFileUrl.length && $simpleLocalAvatarFileUrl.val() !== '' ) {
+		var $simpleDefaultAvatarImg = $simpleLocalAvatarFileUrl.parent().find('img.avatar');
+		$simpleDefaultAvatarImg.attr('src', $simpleLocalAvatarFileUrl.val());
+		$simpleDefaultAvatarImg.attr('srcset', $simpleLocalAvatarFileUrl.val());
+	}
+
+	if ( $simpleLocalAvatarFileId.val() === '' ) {
```
faisal-alvi:
```suggestion
	if ( '' === $simpleLocalAvatarFileId.val() ) {
```

includes/class-simple-local-avatars.php
```diff
@@ -1131,8 +1141,9 @@ public function admin_scripts() {
 			'sla_admin',
 			'slaAdmin',
 			[
-				'nonce' => wp_create_nonce( 'sla_clear_cache_nonce' ),
-				'error' => esc_html__( 'Something went wrong while clearing cache, please try again.', 'simple-local-avatars' ),
+				'nonce'            => wp_create_nonce( 'sla_clear_cache_nonce' ),
+				'error'            => esc_html__( 'Something went wrong while clearing cache, please try again.', 'simple-local-avatars' ),
+				'insertMediaTitle' => __( 'Choose default avatar', 'simple-local-avatars' ),
```
```suggestion
				'insertMediaTitle' => esc_html__( 'Choose default avatar', 'simple-local-avatars' ),
```

includes/class-simple-local-avatars.php
```diff
@@ -1222,7 +1233,44 @@ private function clear_user_avatar_cache( $local_avatars, $user_id, $media_id )
...
+
+		$file_id = filter_input( INPUT_POST, 'simple-local-avatar-file-id', FILTER_SANITIZE_NUMBER_INT );
+
+		// check for uploaded files
+		if ( 'options.php' === $pagenow && ! empty( $file_id ) ) {
+			update_option( 'simple_local_avatar_default', $_POST['simple-local-avatar-file-id'] );
+		}
+	}
+  
+  /** Migrate the user's avatar data from WP User Avatar/ProfilePress
```
faisal-alvi: Looks like a tab space is required.

includes/class-simple-local-avatars.php
```diff
@@ -1222,7 +1233,44 @@ private function clear_user_avatar_cache( $local_avatars, $user_id, $media_id )
...
+	 * @param array $defaults Default options for avatar.
+	 *
+	 * @return array Default options of avatar.
+	 */
+	public function add_avatar_default_field( $defaults ) {
+		if ( ! did_action( 'wp_enqueue_media' ) ) {
+			wp_enqueue_media();
+		}
+		$default_avatar_file_id  = get_option( 'simple_local_avatar_default', '' );
+		$default_avatar_file_url = wp_get_attachment_image_url( $default_avatar_file_id );
```
faisal-alvi: We should check if `$default_avatar_file_id` is not empty. Also, `$default_avatar_file_url` should not be `false` before proceeding.

includes/class-simple-local-avatars.php
```diff
@@ -1222,7 +1233,44 @@ private function clear_user_avatar_cache( $local_avatars, $user_id, $media_id )
...
+		ob_start();
+		?>
+		<input type="hidden" name="simple-local-avatar-file-id" id="simple-local-avatar-file-id" value="<?php echo esc_attr( $default_avatar_file_id ); ?>"/>
+		<input type="hidden" name="simple-local-avatar-file-url" id="simple-local-avatar-file-url" value="<?php echo esc_url( $default_avatar_file_url ); ?>"/>
+		<input type="button" name="simple-local-avatar" id="simple-local-avatar-upload" class="button-secondary" value="<?php esc_html_e( 'Choose Default Avatar', 'simple-local-avatar' ); ?>"/>
```
faisal-alvi: Should we use `esc_attr_e` instead of `esc_html_e`?

includes/class-simple-local-avatars.php
```diff
@@ -1222,7 +1233,44 @@ private function clear_user_avatar_cache( $local_avatars, $user_id, $media_id )
...
+	 * Save default avatar attachment id in option.
+	 */
+	private function save_default_avatar_file_id() {
+		global $pagenow;
+
+		$file_id = filter_input( INPUT_POST, 'simple-local-avatar-file-id', FILTER_SANITIZE_NUMBER_INT );
+
+		// check for uploaded files
+		if ( 'options.php' === $pagenow && ! empty( $file_id ) ) {
+			update_option( 'simple_local_avatar_default', $_POST['simple-local-avatar-file-id'] );
```
```suggestion
			update_option( 'simple_local_avatar_default', $file_id );
```

includes/class-simple-local-avatars.php
```diff
@@ -1222,7 +1233,44 @@ private function clear_user_avatar_cache( $local_avatars, $user_id, $media_id )
...
+	/**
+	 * Save default avatar attachment id in option.
+	 */
+	private function save_default_avatar_file_id() {
+		global $pagenow;
+
+		$file_id = filter_input( INPUT_POST, 'simple-local-avatar-file-id', FILTER_SANITIZE_NUMBER_INT );
+
+		// check for uploaded files
+		if ( 'options.php' === $pagenow && ! empty( $file_id ) ) {
```
faisal-alvi, line 1271: I wonder if `$file_id` could be `empty` as [filter_input](https://www.php.net/manual/en/function.filter-input.php) returns `false` if the filter fails, or `null` if the variable is not set.

includes/class-simple-local-avatars.php
```diff
@@ -1383,5 +1430,5 @@ public function wp_cli_migrate_from_wp_user_avatar( $args, $assoc_args ) {
 				esc_html( $count )
 			)
 		);
-	}
+  }
```
faisal-alvi: Looks like an unnecessary change.

assets/js/admin.js
```diff
@@ -58,4 +61,41 @@
 			},
 		} );
 	}
+
+	$simpleLocalAvatarUpload.click(function(e) {
```
faisal-alvi: Can we have a short documentation/comment on what the code does?

faisal-alvi (changes requested):
@mehulkaklotar thanks for the efforts. I have checked and the avatar seems to be broken after being selected. Not sure if it's an issue from my side.

sla-96
[image link](https://www.screencast.com/t/jbbUMihGM)

Also, I have added a few suggestions/comments below, please have a look.

includes/class-simple-local-avatars.php
mehulkaklotar: @faisal-alvi we have put a check for empty here. if null or empty, it will not update the option. You are suggesting we should put some validation message for user if it fails?

mehulkaklotar:
Hi @faisal-alvi 
Thank you for the great suggestions. I have updated the PR according to your changes. 

About the issue of not appearing the default avatar for you. May be you need to regenerate the assets to see the changes. Please try that and see if that resolves the issue for you.

faisal-alvi (approved):
@mehulkaklotar thank you for your great work! ❤️  🚀

PR #117, opened 2022-05-03
Correct plugin name in changelog

grappler:
[WP User Avatar](https://wordpress.org/plugins/wp-user-avatar/) and [WP User Avatars](https://wordpress.org/plugins/wp-user-avatars/) are different plugins.


jeffpaul:
Nice catch, thanks @grappler!

PR #125, opened 2022-05-09
Fix thumbnail urls for network configured avatars

### LINKED ISSUE #124: Wrong avatar URLs for multisite network configured avatars
### Describe the bug

When local avatars are shared and configured at the network level, the avatar URLs used on the frontend are broken. They end up with the site URL prepended to the absolute path of the avatar thumbnail.

This only happens for non-standard thumbnail sizes (e.g. `100`). Standard sizes (e.g. `96`) work fine.

### Steps to Reproduce

1. Activate the plugin network-wide and activate shared avatars through the `Shared network avatars` options.
2. Go to the network dashboard
3. Go to Users
4. Edit the avatar of a user
5. Go to a post with comments by that user, using a theme that outputs non-standard avatar sizes (like `100`).

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

vladolaru:
### Description of the Change

After generating a non-existent thumbnail size for an avatar in the `Simple_Local_Avatars::get_simple_local_avatar_url()@L336` method, the current logic transforms the destination file path to a URL by replacing the current blog `$upload_path['basedir']` with `$upload_path['baseurl']`. 

This works fine for all situations except when avatars are configured at the network level, since those avatars use the `uploads` directory (`wp-content/uploads`) as a basedir, not the site-specific directory (e.g. `wp-content/uploads/sites/5/`). This is why the destination file path doesn't get converted into a URL, and the plugin ends up caching the full path for that thumbnail size.

My solution is to test for this specific case, and properly transform the path into a URL.

Closes #124 

### Alternate Designs

Not the case.

### Possible Drawbacks

None that I can imagine.

### Verification Process

I have checked both with our themes (@pixelgrade) and with the standard core themes. Since the logic is additional to the existing one, without modifying the existing logic, the changes are safe.

### Checklist:

- [x] All new and existing tests passed.

### Changelog Entry

Fixed - Broken avatar URLs for network-configured shared avatars with non-standard thumbnail sizes.

### Credits

Props @vladolaru

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -358,7 +358,7 @@ class Simple_Local_Avatars {
 							$dest_file_url = '';
 							if ( false !== strpos( $dest_file, $upload_path['basedir'] ) ) {
 								$dest_file_url = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
-							} else if ( is_multisite() && false !== strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
+							} else if ( is_multisite() && strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
 								$dest_file_url = str_replace( ABSPATH . 'wp-content/uploads', network_site_url( '/wp-content/uploads' ), $dest_file );
 							}
 
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -356,7 +356,7 @@ class Simple_Local_Avatars {
 						if ( ! is_wp_error( $saved ) ) {
 							// Transform the destination file path into URL.
 							$dest_file_url = '';
-							if ( false !== strpos( $dest_file, $upload_path['basedir'] ) ) {
+							if ( strpos( $dest_file, $upload_path['basedir'] ) ) {
 								$dest_file_url = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
 							} else if ( is_multisite() && strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
 								$dest_file_url = str_replace( ABSPATH . 'wp-content/uploads', network_site_url( '/wp-content/uploads' ), $dest_file );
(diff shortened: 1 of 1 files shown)


includes/class-simple-local-avatars.php
```diff
@@ -354,7 +354,15 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
 						$dest_file = $editor->generate_filename();
 						$saved     = $editor->save( $dest_file );
 						if ( ! is_wp_error( $saved ) ) {
-							$local_avatars[ $size ] = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
+							// Transform the destination file path into URl.
```
```suggestion
							// Transform the destination file path into URL.
```

includes/class-simple-local-avatars.php
```diff
@@ -354,7 +354,15 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
...
+							$dest_file_url = '';
+							if ( false !== strpos( $dest_file, $upload_path['basedir'] ) ) {
```
faisal-alvi, line 359:
```suggestion
							if ( strpos( $dest_file, $upload_path['basedir'] ) ) {
```

includes/class-simple-local-avatars.php
```diff
@@ -354,7 +354,15 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
...
+								$dest_file_url = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
+							} else if ( is_multisite() && false !== strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
```
faisal-alvi, line 361:
```suggestion
							} else if ( is_multisite() && strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
```

faisal-alvi (changes requested):
@vladolaru Resolves the issue perfectly, thanks for the work! Just minor suggestions added below, then it'll be ready for merge.

vladolaru:
Thanks. I've accepted the suggestions.

vladolaru:
Actually, no :) We need to strictly compare with `false` since `strpos()` returns `0` when the needle is found at the beginning of the string. And `0` evaluates to false, not true when converted to bool.

faisal-alvi (approved):
🚀

PR #127, opened 2022-05-10
Try adding alt text when rendering local avatars

### LINKED ISSUE #126: Ensure alt text is output when the avatar image is output
### Describe the bug

If an author has a local avatar set and the image used for that avatar has custom alt text set, it doesn't appear that we ever use that alt text (see [this report](https://wordpress.org/support/topic/images-dont-get-alt-text)). WordPress already supports outputting the alt text on the avatar image tag if that exists, so we just need to ensure that if someone has a local avatar set, let's grab the alt text for that image and return that in the image arguments.

### Steps to Reproduce

1. Upload a new image and add alt text to it
2. Assign that image as a local avatar to a user
3. Create a post with that same user
4. View the post on the front-end and see the avatar being output but no alt text

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

As mentioned in #126, it doesn't appear the alt text saved with an attachment is ever used when rendering a local avatar. If someone passes in some custom alt text (for instance, using the 4th argument of `get_avatar`) that is then used as the alt text for the image. But if that argument isn't passed in but the attachment we are rendering has alt text assigned to it, we should use that for the alt text argument.

This PR fixes that by adding a check that runs if a local avatar is found AND no custom alt text has been passed in. It then pulls the alt text from the underlying attachment object which will be stored in post meta (if it exists, otherwise just an empty string).

I also added a new method used to get the user ID, to avoid having duplicate code between the existing `get_simple_local_avatar_url` method and the new `get_simple_local_avatar_alt` method. 

Closes #126 

### Alternate Designs

Right now we default to using the alt text that is passed in and then fallback to the alt text assigned to an image. I think that's the right approach but we could flip those, so the alt text saved with an image would take precedence.

### Possible Drawbacks

None

### Verification Process

1. Upload a new image
2. Add alt text to this image
3. Set this image as your local avatar
4. Publish a new post with that user
5. Using a theme that renders an author section (like Twenty Twenty One) view the front-end of that post
6. Inspect the author image and note the alt text being used
7. For a second test, add a call to `get_avatar` and pass in the 4th argument for the alt text
8. Inspect the source again and note the alt text now comes from that function and not the image

### Checklist:

### Changelog Entry

> Added - If an image used for a local avatar has alt text assigned, ensure that alt text is used when rendering the image

### Credits

Props @dkotter, @pixelloop

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -275,7 +275,7 @@ class Simple_Local_Avatars {
 	/**
 	 * Get a user ID from certain possible values.
 	 *
-	 * @since 2.4.1
+	 * @since 2.5.0
 	 *
 	 * @param mixed $id_or_email The Gravatar to retrieve. Accepts a user ID, Gravatar MD5 hash,
 	 *                           user email, WP_User object, WP_Post object, or WP_Comment object.
@@ -397,7 +397,7 @@ class Simple_Local_Avatars {
 	/**
 	 * Get local avatar alt text.
 	 *
-	 * @since 2.4.1
+	 * @since 2.5.0
 	 *
 	 * @param mixed $id_or_email The Gravatar to retrieve. Accepts a user ID, Gravatar MD5 hash,
 	 *                           user email, WP_User object, WP_Post object, or WP_Comment object.
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -262,9 +262,40 @@ class Simple_Local_Avatars {
 
 		if ( ! empty( $args['url'] ) ) {
 			$args['found_avatar'] = true;
+
+			// If custom alt text isn't passed, pull alt text from the local image.
+			if ( empty( $args['alt'] ) ) {
+				$args['alt'] = $this->get_simple_local_avatar_alt( $id_or_email );
+			}
 		}
 
 		return $args;
+	}
+
+	/**
+	 * Get a user ID from certain possible values.
+	 *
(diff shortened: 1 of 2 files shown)


pixelloop:
I have tested the new code for includes/class-simple-local-avatars.php and can confirm that it fixes the alt text issue.

includes/class-simple-local-avatars.php
```diff
@@ -372,6 +394,31 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
 		return esc_url( $local_avatars[ $size ] );
 	}
 
+	/**
+	 * Get local avatar alt text.
+	 *
+	 * @since 2.4.1
```
dkotter:
```suggestion
	 * @since 2.5.0
```

includes/class-simple-local-avatars.php
```diff
@@ -262,21 +262,28 @@ public function get_avatar_data( $args, $id_or_email ) {
...
 
 		return $args;
 	}
 
 	/**
-	 * Get local avatar url.
+	 * Get a user ID from certain possible values.
 	 *
-	 * @since 2.2.0
+	 * @since 2.4.1
```
dkotter:
```suggestion
	 * @since 2.5.0
```

dkotter:
Thanks so much for reporting and testing this @pixelloop!

faisal-alvi (approved):
@dkotter LGTM! 🚀  thanks for the work!

PR #130, opened 2022-05-18
Enhancement/128 ”Skip cropping” + “Free cropping area” added!

### LINKED ISSUE #128: Crop changes: skip cropping; allow non-square crops

In #65 it was requested that we add the ability to choose how an image is cropped (implemented in #83). But there are times when someone uploads an image that is the exact size they need and they should be able to bypass the cropping. Right now, the only option you have is to choose a crop.

I think ideal here would be to have a a `Skip cropping` button, similarly to how the site icon is handled.

Screen Shot 2022-05-13 at 9 45 08 AM

In addition, it seems we force the image to always be a square (which was requested [here](https://github.com/10up/simple-local-avatars/issues/65#issuecomment-912594844)). Is there a reason why we need to force this image to always be a square? Again, using the idea of the site icon, it gives you a pre-defined crop area but than you can change that crop to be whatever size you want. I guess I'd argue we should allow that flexibility instead of forcing a square crop.

### Designs

_No response_

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change

- This PR introduces a new button “Skip cropping”, similarly to how the site icon is handled.
- And, the “Free cropping” area feature for avatar crops.

Closes #128

### Alternate Designs

Screen Shot 2022-05-13 at 9 45 08 AM

### Possible Drawbacks

No more square restrictions in the crop, so the users who want it to be strict "Square", may have to manually crop the image and try to make it square. :|

### Verification Process

1. Build the project `npm run build`.
2. Visit Users > Profile.
3. Click the “Choose from Media Library” button.
4. Select/Upload any image.
5. Click the “Select avatar and Crop” button.
6. Try to change the cropping area to any size and shape.
7. At the bottom right of the popup, see the “Skip cropping” button. Try using it.
8. Also, try cropping various shapes and see if that works.

### Checklist:

### Changelog Entry

> Added - “Skip cropping” button added.
> Added - Free Cropping area introduced, no more strict squares!

### Credits
Props @dkotter @faisal-alvi @cadic

### DIFF
--- a/assets/js/simple-local-avatars.js
+++ b/assets/js/simple-local-avatars.js
@@ -31,31 +31,11 @@ jQuery(document).ready(function ($) {
 		const cropControl = {
 			id: 'control-id',
 			params: {
-				flex_width: false, // set to true if the width of the cropped image can be different to the width defined here
-				flex_height: false, // set to true if the height of the cropped image can be different to the height defined here
+				flex_width: true, // set to true if the width of the cropped image can be different to the width defined here
+				flex_height: true, // set to true if the height of the cropped image can be different to the height defined here
 				width: 200, // set the desired width of the destination image here
 				height: 200, // set the desired height of the destination image here
 			},
-		};
-
-		/**
-		 * Return whether the image must be cropped, based on required dimensions.
-		 *
-		 * @param {boolean} flexW
-		 * @param {boolean} flexH
-		 * @param {number}  distW
-		 * @param {number}  distH
-		 * @param {number}  imgW
-		 * @param {number}  imgH
-		 * @returns {boolean}
-		 */
-		cropControl.mustBeCropped = function (flexW, flexH, distW, distH, imgW, imgH) {
-			// Skip cropping if the image matches the crop dimension.
-			if (imgW === distW && imgH === distH) {
-				return false;
-			}
-
-			return true;
 		};
 
 		/**
@@ -354,10 +334,8 @@ function simple_local_avatar_calculate_image_select_options(attachment, controll
 
 	const ratio = xInit / yInit;
 
-	controller.set(
-		'canSkipCrop',
-		!control.mustBeCropped(false, false, xInit, yInit, realWidth, realHeight),
-	);
+    // Enable skip cropping button.
+    controller.set('canSkipCrop', true);
 
 	const xImg = xInit;
 	const yImg = yInit;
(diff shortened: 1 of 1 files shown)


assets/js/simple-local-avatars.js
```diff
@@ -354,10 +334,8 @@ function simple_local_avatar_calculate_image_select_options(attachment, controll
 
 	const ratio = xInit / yInit;
 
-	controller.set(
-		'canSkipCrop',
-		!control.mustBeCropped(false, false, xInit, yInit, realWidth, realHeight),
-	);
+    // Enable skip cropping button.
```
cadic: Looks like incorrect indentation here

cadic (dismissed):
Code looks good, except one small indent problem

dkotter:
I know this has been merged already but is it worth discussing this Possible Drawbacks point:

> No more square restrictions in the crop, so the users who want it to be strict "Square", may have to manually crop the image and try to make it square

In the issue I opened, I mentioned that it seemed like we should allow crops other than squares but that is only my opinion. I'm assuming the decision was made initially to force this to a square for a particular reason, so just want to see if there's any other feedback on that before this is shipped? I think for sure the "skip crop" option is needed but there may be a good reason to keep the square size restriction

jeffpaul:
Any chance that it supports a shift+click function that locks the crop ratio to a square?  I agree that it would be nice to keep it "simple" for someone to crop in a square while still allowing for non-square crops for whatever reason (like the one in the original dotorg report).

faisal-alvi:
Agree, I will check about `shift+click` or any other possibilities, otherwise, we might have to add a custom button for square shape.

An example screenshot from https://www.fotor.com/features/crop,

image

cadic:
Maybe there is a change to overload the `onSelectChange` event of `imgAreaSelect` jQuery plugin which is used to crop attachments?

https://github.com/WordPress/WordPress/blob/766a38f2fb420aafb9049ffe43e4a04f35a0c7c8/wp-admin/js/image-edit.js#L786-L800

victorberland:
Is there an ETA for when this will be released?

faisal-alvi:
@victorberland this is on hold to be released until #131 is implemented because as per discussion in https://github.com/10up/simple-local-avatars/pull/130#issuecomment-1132219294, we have to consider the following possible drawback. So probably, this would be implemented by the end of this month and released the early next month. 🤞🏻 

> No more square restrictions in the crop, so the users who want it to be strict "Square", may have to manually crop the image and try to make it square

faisal-alvi:
> Is there an ETA for when this will be released?

@victorberland we are glad to inform you that we have a new release `v2.5.0` which contains the "Skip cropping" button from this PR.

https://wordpress.org/plugins/simple-local-avatars/

PR #133, opened 2022-05-26
bbPress - make sure the button to remove the avatar works

foliovision:
### Description of the Change

When the user avatar is stored, the "Delete local avatar" button on bbPress edit page does not work.

The button is actually a simple link with the `user_id={ID}` and and `action=remove-simple-local-avatar`.

This improvement makes sure it's processed for bbPress edit profile page.

### Alternate Designs

It might be better to adjust admin_enqueue_scripts() to work on bbPress edit profile page, it would also use the nonce that way.

### Possible Drawbacks

It's missing the nonce security.

### Verification Process

The "Delete local avatar" on the bbPress profile editing is now able to delete the stored avatar.

### Checklist:

### Changelog Entry

Fixed - avatar removal on bbPress profile edit page

### Credits

Props @foliovision

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -118,6 +118,8 @@ class Simple_Local_Avatars {
 		add_action( 'wp_ajax_remove_simple_local_avatar', array( $this, 'action_remove_simple_local_avatar' ) );
 		add_action( 'user_edit_form_tag', array( $this, 'user_edit_form_tag' ) );
 
+		add_action( 'init', array( $this, 'front_end_remove_avatar' ) );
+
 		add_action( 'rest_api_init', array( $this, 'register_rest_fields' ) );
 
 		add_action( 'wp_ajax_migrate_from_wp_user_avatar', array( $this, 'ajax_migrate_from_wp_user_avatar' ) );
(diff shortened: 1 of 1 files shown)


faisal-alvi (changes requested):
@foliovision thanks for the work here. However, IMO we can adopt another more robust solution here.

> It might be better to adjust admin_enqueue_scripts() to work on bbPress edit profile page, it would also use the nonce that way.

Agree with this statement. The bbPress loads the same "edit profile" template on the FE.

image
[image link](https://www.screencast.com/t/ONhUmRb3)

But currently, the SLA plugin does not perform any JS actions at the FE side which is causing the issue here. So instead of handling the page load action (i.e. additional code), we should move the "Delete action" JS code to a separate file and load that file on both sides (BE & FE).

@jeffpaul @dkotter any thoughts on this?

dkotter:
Is the only thing not working here the delete portion? And is the code we need to handle this the exact same as the code we use on the admin side? If so, I'm fine with loading that on the FE if on a bbPress edit profile page.

Any downside to just loading the entire JS file instead of having to abstract portions out? Seems like there's other things we'll need as well, like the user ID and nonce, where we could see if bbPress fires an action when on an edit profile page and hook in our same `admin_enqueue_scripts` method.

foliovision:
I agree it would be better to include all the same scripts as when editing the profile in wp-admin and then the same Ajax should work.

Thanks,
Martin

faisal-alvi:
@dkotter 
> Is the only thing not working here the delete portion?

so far, yes.


> And is the code we need to handle this the exact same as the code we use on the admin side? If so, I'm fine with loading that on the FE if on a bbPress edit profile page.

yes, the code is same.

> Any downside to just loading the entire JS file instead of having to abstract portions out? Seems like there's other things we'll need as well, like the user ID and nonce, where we could see if bbPress fires an action when on an edit profile page and hook in our same admin_enqueue_scripts method.

We can load the entire JS file only on the bbPress edit profile page but most of the JS code will be unnecessarily loaded as it is required only for the BE side.

faisal-alvi:
Closing after #134.

PR #132, opened 2022-05-26
bbPress - make sure the functions to process image uploads are all there

foliovision:
### Description of the Change

When user tries to submit the image on the bbPress profile editing page, it fails to process. Some functions are missing and user gets a white screen.

### Alternate Designs

None.

### Possible Drawbacks

None.

### Verification Process

bbPress user profile editing now saves properly when the image is selected for avatar.

### Checklist:

- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

### Changelog Entry

Fixed - image upload on bbPress profile edit page

### Credits

Props @foliovision

PR #134, opened 2022-06-03
Enhancement/bbPress The support for the bbPress plugin!

faisal-alvi:
### Description of the Change

- This PR enhances the SLA plugin to support the bbPress plugin.
- Loads the entire JS at bbPress edit profile page at FE to support 1) the SLA avatar upload 2) Remove the SLA avatar.
- Also, forces the `file` input field at FE because we don’t have a media popup at FE.
- Adds support for the file uploads from FE by loading the required core files for `wp_handle_upload` and `wp_generate_attachment_metadata` functions. Thanks to @foliovision to contribute in this point via #132.

Enhancement of the #132 and Alternate of #133.

### Alternate Designs

- The file input field to upload SLA are now visible at FE on the edit bbPress profile page.

### Possible Drawbacks

Nothing seems to be.

### Verification Process

1. Install the [bbPress](https://wordpress.org/plugins/bbpress/) plugin.
2. Visit https://site.com/forums/users/{user-name}/edit
3. See the file input displayed to upload the avatar.
4. Upload an avatar and confirm it's uploaded successfully.
5. Click the “Delete local avatar” button and see if the avatar was deleted successfully.
6. Also check these actions work in the backend as well.

### Checklist:

### Changelog Entry

> Added - The support for the bbPress plugin by loading the JS at FE on the profile edit page.
> Fixed - Uploads and Deletions work at FE on the bbPress edit profile page.

### Credits

Props @foliovision @faisal-alvi @iamdharmesh

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -846,7 +846,7 @@ class Simple_Local_Avatars {
 							// At FE, show the file input field regardless of the caps.
 							if ( ! is_admin() || ! current_user_can( 'upload_files' ) ) {
 								?>
-								<p style="display: inline-block;">
+								<p style="display: inline-block; width: 26em;">
 									<span class="description"><?php esc_html_e( 'Choose an image from your computer:' ); ?></span><br />
 									<input type="file" name="simple-local-avatar" id="simple-local-avatar" class="standard-text" />
 									<span class="spinner" id="simple-local-avatar-spinner"></span>
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -687,7 +687,7 @@ class Simple_Local_Avatars {
 		 * @param string $hook_suffix Page hook.
 		 * @return array
 		 */
-		$screens = apply_filters( 'simple_local_avatars_enqueue_scripts', array( 'profile.php', 'user-edit.php', 'options-discussion.php' ), $hook_suffix );
+		$screens = apply_filters( 'simple_local_avatars_admin_enqueue_scripts', array( 'profile.php', 'user-edit.php', 'options-discussion.php' ), $hook_suffix );
 
 		// Allow SLA actions on a bbPress profile edit page at FE.
 		if ( function_exists( 'bbp_is_user_home_edit' ) && bbp_is_user_home_edit() ) {
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -107,7 +107,10 @@ class Simple_Local_Avatars {
 
 		add_action( 'admin_init', array( $this, 'admin_init' ) );
 
-		add_action( 'admin_enqueue_scripts', array( $this, 'admin_enqueue_scripts' ) );
+		// Load the JS on BE & FE both, in order to support third party plugins like bbPress.
+		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
+		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
+
 		add_action( 'show_user_profile', array( $this, 'edit_user_profile' ) );
 		add_action( 'edit_user_profile', array( $this, 'edit_user_profile' ) );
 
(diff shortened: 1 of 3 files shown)


foliovision:
Hello @faisal-alvi 

perfect, it works now. I was able to add and remove the avatar on a bbPress edit profile page without issues.

Thanks,
Martin

includes/class-simple-local-avatars.php
```diff
@@ -107,7 +107,10 @@ public function add_hooks() {
 
 		add_action( 'admin_init', array( $this, 'admin_init' ) );
 
-		add_action( 'admin_enqueue_scripts', array( $this, 'admin_enqueue_scripts' ) );
+		// Load the JS on BE & FE both, in order to support third party plugins like bbPress.
+		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
```
iamdharmesh, line 111: Do you think we should only enqueue scripts in FE if the bbPress plugin is active?

includes/class-simple-local-avatars.php
```diff
@@ -684,7 +687,12 @@ public function admin_enqueue_scripts( $hook_suffix ) {
 		 * @param string $hook_suffix Page hook.
 		 * @return array
 		 */
-		$screens = apply_filters( 'simple_local_avatars_admin_enqueue_scripts', array( 'profile.php', 'user-edit.php', 'options-discussion.php' ), $hook_suffix );
```
iamdharmesh, line 687: I think we should keep this as it is because some users might be already using it and if we want to change it then better to add a changelog about the rename of the filter.

includes/class-simple-local-avatars.php
```diff
@@ -684,7 +687,12 @@ public function admin_enqueue_scripts( $hook_suffix ) {
...
+		$screens = apply_filters( 'simple_local_avatars_enqueue_scripts', array( 'profile.php', 'user-edit.php', 'options-discussion.php' ), $hook_suffix );
+
+		// Allow SLA actions on a bbPress profile edit page at FE.
+		if ( function_exists( 'bbp_is_user_home_edit' ) && bbp_is_user_home_edit() ) {
+			$hook_suffix = 'profile.php';
+		}
 
 		if ( ! in_array( $hook_suffix, $screens, true ) ) {
```
iamdharmesh: Better to modify the condition here something like below instead of mocking **$hook_suffix** on line 694. what do you think?
`if ( ! in_array( $hook_suffix, $screens, true ) || ! (function_exists( 'bbp_is_user_home_edit' ) && bbp_is_user_home_edit()) ) {`

iamdharmesh (commented):
Thanks for the PR @faisal-alvi. The code looks good to me, just added a few notes on this, can you please help to check on it?

includes/class-simple-local-avatars.php
faisal-alvi: It's a good suggestion but we are already checking the [screen](https://github.com/10up/simple-local-avatars/pull/134/files/6a939c917a5973acaf7d3626ad3c99718743637e#diff-a05d559d485d184224ad87c7e310a12068ce060290bcdda4e1001d3d4e6bea89R690) inside the `enqueue_scripts()` so avoided double-checking.

includes/class-simple-local-avatars.php
faisal-alvi: thanks for bringing this to my attention, I have just noticed the `$hook_suffix` is a parameter coming with the [admin_enqueue_scripts](https://developer.wordpress.org/reference/hooks/admin_enqueue_scripts/) only, not with [wp_enqueue_scripts](https://developer.wordpress.org/reference/hooks/wp_enqueue_scripts/). So we can keep the filter as is because it will only be applicable to the BE side. Probably we can introduce a new filter to load JS on specific pages at FE in the future. For now, reverting it back.

includes/class-simple-local-avatars.php
faisal-alvi: just wanted to keep the code cleaner, and nothing to worry about overwriting the `$hook_suffix` as it will always be `false` at FE.

iamdharmesh (approved):

Issue #131, opened 2022-05-25
Add: Crop area options: Square, Freeform


faisal-alvi:
### Is your enhancement related to a problem? Please describe.

* The crop feature was introduced in #83 with forced Square shape selection.
* Later, a "Skip crop" button and "Freeform" shape are added in #130 but now the Square shape selection is removed.
* So it is discussed from https://github.com/10up/simple-local-avatars/pull/130#issuecomment-1132219294 that we should allow users to crop in a "Square" shape or a "Freeform".

### Designs

Design discussion for the following area will be required, (i.e. **a:** where should we put buttons and **b:** how buttons should look like):

<img width="1274" alt="image" src="https://user-images.githubusercontent.com/25176325/170189260-7a2d65b6-b25c-4e8e-b49f-006adc7943af.png">

Go to "Users > Profile" (your edit profile section), and click the "Select avatar and crop" button to see the above screen.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jeffpaul:
Initial thought is left aligned buttons for `Square Crop` (default) and `Freeform Crop`, though maybe they're not buttons so much as a left/right toggle for square vs. freeform crop since it will be one-or-the-other?

faisal-alvi:
@dinhtungdu as discussed in the meeting, your point is that allowing a "freeform" crop does not make sense/worth because the end result of the avatar being loaded at FE is also cropped by the [WordPress](https://developer.wordpress.org/reference/functions/get_avatar/). So should we stick to the "strict square" selection?

Even setting the avatar with the "freeform" selection:

![image](https://user-images.githubusercontent.com/25176325/171826638-a19564ed-e960-4f22-83b9-e4635739e9e5.png)

At FE, the avatar loads cropped:

![image](https://user-images.githubusercontent.com/25176325/171826807-88df5654-d40a-4461-9dab-d699307be94f.png)

Please correct me if this was not the exact point.

dinhtungdu:
@faisal-alvi you said it correctly! Freeform cropping creates more confusion than it solves IMO. If the final avatar displayed on the front end is still a square one. Even if we can hack it by storing the free-form cropped version in the user metadata, we still have a problem with the output HTML of the `get_avatar()` function, which set both `width` and `height` to the same value for the `<img>` tag.

faisal-alvi:
@dkotter any thoughts on Tung's ^comment?

dinhtungdu:
[...]

PR #136, opened 2022-06-17
Fix/135 Re-adding "Square" Crop ability

### LINKED ISSUE #135: Re-add "Square" crop and remove "Freeform"
### Describe the bug

As a conclusion of the following discussions,
* https://github.com/10up/simple-local-avatars/pull/130#issuecomment-1132219294
* And from https://github.com/10up/simple-local-avatars/issues/131#issuecomment-1145771123 to https://github.com/10up/simple-local-avatars/issues/131#issuecomment-1155913693

It is concluded that we should remove the freeform crop ability (and re-add the square crop ability) and we can consider that in the future if we get more user requests.

### Steps to Reproduce

1. Edit your profile.
2. Try to set the avatar.
3. You can crop the avatar in any shape (freeform ability), it's not strict to the square shape.

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change

* This PR removes the "freeform" ability and re-adds the "sure" crop ability (which was removed in #130).
* Also updates the button name from "Skip Crop" to "Default Crop" only on the edit profile page.

Closes #135

### Alternate Designs

n/a

### Possible Drawbacks

Maybe in the future, we get some users wanting the freeform crop ability.

### Verification Process

1. Edit your profile.
2. Set an avatar.
3. See, you are forced to stick with the Square form while cropping an avatar.
4. Also check if the button name from "Skip Crop" to "Default Crop" is only changed on the edit profile page, not any other page like when selecting the site icon in the customizer (`wp-admin/customize.php?return=%2Fwp-admin%2Fprofile.php`), the button name should be "Skip Crop".

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.

### Changelog Entry

> Added - Re-added the Square crop ability.

### Credits

Props @dkotter @dinhtungdu @faisal-alvi

### DIFF
--- a/assets/js/simple-local-avatars.js
+++ b/assets/js/simple-local-avatars.js
@@ -364,6 +364,7 @@ function simple_local_avatar_calculate_image_select_options(attachment, controll
 		y1,
 		x2: xInit + x1,
 		y2: yInit + y1,
+		aspectRatio: `${xInit}:${yInit}`,
 	};
 }
 
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -99,6 +99,7 @@ class Simple_Local_Avatars {
 	 * Register actions and filters.
 	 */
 	public function add_hooks() {
+		global $pagenow;
 
 		add_filter( 'plugin_action_links_' . SLA_PLUGIN_BASENAME, array( $this, 'plugin_filter_action_links' ) );
 
@@ -133,6 +134,14 @@ class Simple_Local_Avatars {
 
 		add_filter( 'avatar_defaults', array( $this, 'add_avatar_default_field' ) );
 		add_action( 'wpmu_new_blog', array( $this, 'set_defaults' ) );
+
+		if ( 'profile.php' === $pagenow ) {
+			add_filter( 'media_view_strings', function ( $strings ) {
+				$strings['skipCropping'] = esc_html__( 'Default Crop', 'simple-local-avatars' );
+
+				return $strings;
+			}, 10, 1 );
+		}
 	}
 
 	/**


peterwilsoncc (dismissed):
LGTM.

The skip cropping button remains in place (which actually uses the default square crop during testing), was that intentional? I am presuming it is and approving the change for now.

faisal-alvi:
@peterwilsoncc Thanks for raising this point, this seems similar mentioned by @dinhtungdu in https://github.com/10up/simple-local-avatars/issues/131#issuecomment-1147014435. Even if a user skips the crop, the end result will still be in the (default) square.

WordPress displays an avatar in its appropriate/nearest scaled size at FE, we are also using the same avatar size that is found in the arguments passed to `get_avatar_data()` and displaying our Simple Local Avatar (regardless of the avatar was skipped for crop or not). 

https://github.com/10up/simple-local-avatars/blob/a6d3af85506592531cc773c4a144ebd0e52074e9/includes/class-simple-local-avatars.php#L256

If a user crops avatar, a **cropped image's scaled version** is displayed at FE "vs", If cropping is skipped, the **original image's scaled version** is displayed. In either case, the scaled (or we can say "cropped") image is displayed at FE. This is WP's default behavior.

Now the question is - are we supposed to display an 'original' size avatar when the user clicks "Skip Cropping"? In that case, we have to add a new user meta that stores whether the avatar is cropped or not. And conditionally skip passing the size in the above code and use something like this:

` $simple_local_avatar_url = $this->get_simple_local_avatar_url( $id_or_email, 0 );` (where 0 will bring the original image URL)
 
I have uploaded and set a 460 x 6358 (vertical) avatar image and used the 'original' version (by using the above code statically), and here are the results.

image

IMO, we should not change the WP's default behavior of displaying avatars at the front end (by displaying 'original' sized avatars) and keep using the "default" crop/scaled version of Avatar. However, I'm concerned users may get confused about why the avatar is in default cropped mode even if they have clicked "Skip Crop".

@dkotter @jeffpaul - please share your thoughts, is it fine to keep using the scaled avatar images by WordPress at the front end (even when the user chooses the "Skip cropping" while setting up an avatar in the backend?

peterwilsoncc:
> IMO, we should not change the WP's default behavior of displaying avatars at the front end (by displaying 'original' sized avatars) and keep using the "default" crop/scaled version of Avatar. However, I'm concerned users may get confused about why the avatar is in default cropped mode even if they have clicked "Skip Crop".

This makes sense.

How about instead of "Skip Crop" the text is changed to "Default Crop". This can be discussed on another ticket as this PR covers a different change.

faisal-alvi:
@peterwilsoncc the "Skip Crop" is coming from the core and I didn't find a way to overwrite it.

https://github.com/WordPress/WordPress/blob/9e3dc52d45f6023df554ab421f9daf34f2383444/wp-includes/media.php#L4600

**EDIT:** Found [the filter](https://github.com/WordPress/WordPress/blob/9e3dc52d45f6023df554ab421f9daf34f2383444/wp-includes/media.php#L4665) `media_view_strings` just below that code to modify the strings.

peterwilsoncc (approved):
Looks good, thanks.

PR #143, opened 2022-07-22
Enhancement/140 Bump WordPress and PHP minimums

### LINKED ISSUE #140: Bump WordPress and PHP minimums
Bump WP & PHP Minimums to the following
 
- [x] WordPress to 5.7
- [x] PHP to 7.4
- [x] Remove any no-longer-needed conditional code for older version support.
- [x] Take a look at adjusting our testing matrices, if applicable.
- [x] Bump the minimums in plugin documentation and header fields.

faisal-alvi:
### Description of the Change
Bumped the minimum WordPress and PHP minimums.
* WordPress to 5.7
* PHP to 7.4

Closes #140.

### How to test the Change
* Check if the plugin is working as expected.
* Check the Github Actions working without fails.

### Credits

Props @vikrampm1 @faisal-alvi @cadic 

### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,10 +31,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # We claim to support from 5.6+ but WP Mock only supports 7.1+
-        # Also an issue with WP Mock not working on PHPUnit 9.5+, which PHP 7.3+ needs.
-        # php: [ '5.6', '7.0', '7.4', '8.0', '8.1' ]
-        php: [ '7.1', '7.2' ]
+        php: [ '7.4', '8.0', '8.1' ]
 
     steps:
     - name: Checkout
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,28 +10,13 @@ on:
       - develop
 
 jobs:
-#  test_wpa:
-#    runs-on: ubuntu-latest
-#
-#    steps:
-#    - name: Checkout
-#      uses: actions/checkout@v2
-#    - name: Set PHP version
-#      uses: shivammathur/setup-php@v1
-#      with:
-#        php-version: '7.2'
-#        coverage: none
-#
-#    - name: Install dependencies
-#      run: composer install
-
   phpunit:
     name: ${{ matrix.php }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        php: [ '7.1', '7.2' ]
+        php: [ '7.4', '8.0' ]
 
     steps:
     - name: Checkout
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [ '7.4', '8.0', '8.1' ]
+        php: [ '7.1', '7.2' ]
 
     steps:
     - name: Checkout
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "source": "https://github.com/10up/simple-local-avatars"
   },
   "require": {
-    "php": ">=7.4"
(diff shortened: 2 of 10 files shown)


faisal-alvi:
The PHP Unit tests are failing with PHPUnit `9.5.21` version. Hence I have changed the min required PHP in the composer file to `7.1`, which keeps PHPUnit `7.5.20` and the tests run successfully.
```
"require": {
    "php": ">=7.1"
  },
```
@cadic I saw you mentioned about this somewhere recently, any suggestions here would be helpful.

composer.json
```diff
@@ -19,7 +19,7 @@
     "source": "https://github.com/10up/simple-local-avatars"
   },
   "require": {
-    "php": ">=5.6"
+    "php": ">=7.1"
```
cadic: Should we put 7.4 here?

.github/workflows/test.yml
```diff
@@ -31,9 +31,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # We claim to support from 5.6+ but WP Mock only supports 7.1+
-        # Also an issue with WP Mock not working on PHPUnit 9.5+, which PHP 7.3+ needs.
-        # php: [ '5.6', '7.0', '7.4', '8.0', '8.1' ]
         php: [ '7.1', '7.2' ]
```
cadic: 7.1 and 7.2 looks like a deprecated support, needs to be updated to 7.4 and 8.0?

cadic (changes requested):
@faisal-alvi thank you for this update, I left couple questions about PHP versions which are still below the current minimum.

composer.json
faisal-alvi: I had put `7.4` but as mentioned in the [comment above](https://github.com/10up/simple-local-avatars/pull/143#issuecomment-1192505559), it upgrades the PHPUnit to `9.5.21` and all tests [failed](https://github.com/10up/simple-local-avatars/runs/7467741053?check_suite_focus=true). Hence I reverted it back to `7.1` (https://github.com/10up/simple-local-avatars/pull/143/commits/0712ba02be6a580f8016e2fadf4b7d12f70cd7df) that fixed this issue. Any suggestions?

.github/workflows/test.yml
faisal-alvi: Yes, we should, but need to fix the PHPUnit test failure issue first as mentioned above.

faisal-alvi: Done, https://github.com/10up/simple-local-avatars/pull/143/commits/c46a7337186d93b34c21358400f48b70fd8e8a5c

composer.json
faisal-alvi: I noticed that the PHPUnit failure was due to the fix required in the upstream and the PR is already raised https://github.com/10up/wp_mock/pull/164.

Upgraded to `7.4` (https://github.com/10up/simple-local-avatars/pull/143/commits/c46a7337186d93b34c21358400f48b70fd8e8a5c)

faisal-alvi:
As mentioned in https://github.com/10up/simple-local-avatars/pull/143#discussion_r942310144, the PHPUnit failure is due to the fix required in the upstream, and the PR is already raised https://github.com/10up/wp_mock/pull/164, so we can ignore the PHPUnit failures in the Github action(s).

composer.json
cadic: getAnnotations() error is fixed by changing `parent::setUp()` to `\WP_Mock::setUp()` in the test declarations https://github.com/10up/simple-local-avatars/pull/143/commits/2b8e16054e6cf9322728cd2fe8aaff0de04d7231

But there are still 3 tests fail after this fix
> Tests: 30, Assertions: 270, Errors: 1, Failures: 2.

cadic (changes requested):
Please check 3 tests are still failing after fixing the getAnnotation error

cadic (approved):
LGTM!

PR #145, opened 2022-08-05
Fix/144 Update the release process

### LINKED ISSUE #144: Bug: Composer package not functional since 2.3.0
### Describe the bug

Hello! Looks like version 2.3.0 and upwards no longer bundles built/dist files, which makes the Composer package not usable without an additional build/install step, for which pulling down the dev dependencies would be required.

Would it be possible to include the dist files when new versions are tagged, similar to how https://github.com/10up/ElasticPress/issues/2727 was addressed?

Thank you!

### Steps to Reproduce

1. Run: `composer require 10up/simple-local-avatars`
2. Add: `require_once $path_to_vendor . '/10up/simple-local-avatars/simple-local-avatars.php';`
3. Browse to user profile in wp-admin
4. See Console/Network for 404 error on url/to/vendor/10up/simple-local-avatars/dist/simple-local-avatars.js

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change

This PR changes the release process so we can distribute the plugin via composer without needing WPackagist.

Closes #144

### How to test the Change

Test it in the next release (2.6.0).

### Changelog Entry
Changed: The plugin is now available via Composer without any additional steps required.

### Credits

Props @faisal-alvi @kovshenin @jeffpaul 

### Checklist:

### DIFF
--- a/.github/workflows/build-and-tag.yml
+++ b/.github/workflows/build-and-tag.yml
@@ -1,0 +1,35 @@
+name: Build and Tag
+on:
+  push:
+    branches:
+      - 'trunk'
+
+jobs:
+  wordpress:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Install Dependencies
+        run: npm install
+
+      - name: Build Assets
+        run: npm run build
+
+      - name: Setup
+        run: 'echo "VERSION=$(grep -Po ''\"version\": \"[0-9\\.]+\"'' package.json | grep -Po ''[0-9\\.]+'')" >> $GITHUB_ENV'
+
+      - name: Tag
+        run: |
+          echo "Releasing version $VERSION ..."
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git checkout -b "release-$VERSION"
+          git add -f dist/*
+          git commit -m "Release $VERSION"
+          git tag "$VERSION"
+          git push --tags
+        env:
+          TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/build-and-tag.yml
```diff
@@ -0,0 +1,35 @@
+name: Build and Tag
+on:
+  push:
+    branches:
+      - 'trunk'
```
jeffpaul, line 5: There have been _some_ cases where I've had to push to `trunk` twice if my local happened to somehow to be out of sync with `develop`, so pivoting this off pushes to `trunk` worries me a little but hopefully those are extreme edge cases (and perhaps I should do a better job ensuring I'm in sync).

jeffpaul (commented):
Curious why not just update https://github.com/10up/simple-local-avatars/blob/703ab09a9e45e1d0e9bb3da07c2d7b09db418d37/.github/workflows/push-deploy.yml#L2-L5 to be on a GitHub publish and adding in the ZIP file generation to the end?

.github/workflows/build-and-tag.yml
faisal-alvi: Does this [discussion](https://github.com/10up/ElasticPress/pull/2799/files#r881829479) relate (and also answer) to your concern? if not, we should brainstorm on the solution.

faisal-alvi:
@jeffpaul we have already used the Github publish as seen in the [ElasitcPress](https://github.com/10up/ElasticPress/blob/6ed9cca461abaaa9c716efbc8f9a13c213e59da1/.github/workflows/push-deploy.yml).

https://github.com/10up/simple-local-avatars/blob/a179844e2f0fbb1a741b5417905b39bd0e9d7856/.github/workflows/push-deploy.yml#L2-L4

Please let me know if this is not what you are trying to point out.

jeffpaul:
@faisal-alvi yes, I think mimicking EP and releasing when we publish on GH (vs. just on tagging) makes most sense to me; please update there and then I think this is good to merge.

jeffpaul (approved):
Looks good to me, will see how the next release goes, thanks!

PR #147, opened 2022-08-26
Ensure default avatars get alt text

### LINKED ISSUE #146: Ensure alt text is output when using a default avatar
### Describe the bug

Similar to the issue reported in #126 (and fixed in #127), it was [reported](https://wordpress.org/support/topic/images-dont-get-alt-text/#post-15933734) that if you're using a default avatar, that still won't get alt text.

### Steps to Reproduce

1. With Simple Local Avatars active, under the Discussion settings, ensure avatars are turned on and choose a default avatar to use
2. Ensure you're using a theme that outputs avatar images, go to a post written by an author that does not have a custom avatar set
3. You should see the default avatar being output
4. Inspect the source of that image and you should see no alt text there

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

In #127 we ensured that if an image that was set as an avatar had alt text associated with it, that alt text would be output. But if someone doesn't have a custom avatar set and instead we use a default avatar, that still doesn't output alt text.

This PR fixes that by:

1. Setting default alt text that will be used if no alt text can be found (defaulted to `Avatar photo` with a new filter around that so it can be changed)
2. If the default avatar being used was uploaded (instead of a default gravatar image) we try and pull the alt text from that image, if it exists. If not, we use the default mentioned above

Closes #146 

### How to test the Change

1. Ensure you're using a theme that outputs avatars
2. Set a default avatar to be one of the Gravatar options
3. Go to a post that was authored by someone without a custom avatar set
4. Inspect the source of the avatar and note the alt text should be showing `Avatar photo`
5. Change the default avatar to be an uploaded image that has alt text set
6. Inspect the source again and note the alt text should now be what was set on that image
7. If desired, test the `simple_local_avatars_default_alt` filter to ensure you can change the default alt text

### Changelog Entry

> Added - If a default avatar image is used, ensure that outputs alt text. This will either be default text (`Avatar photo`) or the alt text from the uploaded default image

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -426,17 +426,36 @@ class Simple_Local_Avatars {
 	public function get_simple_local_avatar_alt( $id_or_email ) {
 		$user_id = $this->get_user_id( $id_or_email );
 
+		/**
+		 * Filter the default avatar alt text.
+		 *
+		 * @param string $alt Default alt text.
+		 * @return string
+		 */
+		$default_alt = apply_filters( 'simple_local_avatars_default_alt', __( 'Avatar photo', 'simple-local-avatars' ) );
+
 		if ( empty( $user_id ) ) {
-			return '';
+			return $default_alt;
 		}
 
 		// Fetch local avatar from meta and make sure we have a media ID.
 		$local_avatars = get_user_meta( $user_id, 'simple_local_avatar', true );
 		if ( empty( $local_avatars['media_id'] ) ) {
-			return '';
+			$alt = '';
+			// If no avatar is set, check if we are using a default avatar with alt text.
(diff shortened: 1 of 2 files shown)


faisal-alvi (approved):
- [x] ✅ Code LGTM
- [x] ✅ Working as expected.
- [x] ✅ Issue re-generated in the `develop` branch:

image
[image link](https://www.screencast.com/t/EsstKaJVaZH)

- [x] ✅ Default alt text:

image
[image link](https://www.screencast.com/t/DqXJCeswh8a)

- [x] ✅ Alt text from image:

image
[image link](https://www.screencast.com/t/ioGcBAEK)

PR #149, opened 2022-08-31
Enhancement/148 | Expose hooks for "avatar updated" and "avatar deleted"

### LINKED ISSUE #148: Add hooks for "avatar updated" and "avatar deleted", to make SLA more "pluggable" in the WP ecosystem.

I have been using this plugin for awhile and love it's functionality. However, an issue came up recently when I was trying to share the local avatar with an outside service that sits next to Wordpress in our larger application context.

Specifically, I needed to share this avatar with the remote service whenever it changed, or when it was deleted (to clear the remote avatar back to default).

Generally, this feature is also useful for any developers who are implementing their own "edit profile" (and edit avatar) functionality outside of the Wordpress in-admin flow. Really, the use cases are enumerable. Exposing SLA's core events as hooks serves to help SLA "play nice with others".

### Designs

_No response_

I have tried as many ways as I could think of to achieve this with existing wordpress hooks and filters surrounding profile updates and avatars. However, none of these attempts was "bulletproof". In one case, I thought I had a working solution, but Simple Local Avatar's caching functionality was causing false-positives to fire on page-load, so I was sharing the avatar with the remote on every page load.

I wish I could be more specific here on each thing that I tried, but since my final "bulletproof" solution involved forking the SLA plugin and adding these hooks, I have lost those previous attempts to the ether.

Shortly, I will submit a simple PR which adds this functionality. If accepted, documentation would need to be added as well.
**EDIT:** PR: #149 

### Code of Conduct

t-lock:
### Description of the Change

Exposes two hooks, `simple_local_avatar_updated` and `simple_local_avatar_deleted`. These hooks allow theme developers or other plugin developers to react to changes in local avatars in a consistent and precise way. 

Example use cases include:
- sharing user avatars with external services such as chat or forum add-ons
- developing front-end flows for editing user profiles
- any use case adjacent to user management

Closes #148 

### Credits
@t-lock

### Checklist:

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -953,6 +953,14 @@ class Simple_Local_Avatars {
 		$meta_value['blog_id'] = get_current_blog_id();
 
 		update_user_meta( $user_id, $this->user_key, $meta_value ); // save user information (overwriting old).
+
+		/**
+		 * Enable themes and other plugins to react to changes to a user's avatar.
+		 *
+		 * @since 2.6.0
+		 *
+		 * @param int $user_id Id of the user who's avatar was updated
+		 */
 		do_action( 'simple_local_avatar_updated' , $user_id );
 	}
 
@@ -1055,6 +1063,14 @@ class Simple_Local_Avatars {
 			}
 
 			$this->avatar_delete( $user_id );    // delete old images if successful
+
+			/**
+			 * Enable themes and other plugins to react to avatar deletions.
+			 *
+			 * @since 2.6.0
+			 *
+			 * @param int $user_id Id of the user who's avatar was deleted.
+			 */
 			do_action( 'simple_local_avatar_deleted', $user_id );
 
 			if ( defined( 'DOING_AJAX' ) && DOING_AJAX ) {
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -953,7 +953,7 @@ class Simple_Local_Avatars {
 		$meta_value['blog_id'] = get_current_blog_id();
 
 		update_user_meta( $user_id, $this->user_key, $meta_value ); // save user information (overwriting old).
-		do_action("simple_local_avatar_updated", $user_id);
+		do_action( 'simple_local_avatar_updated' , $user_id );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


t-lock:
I have not written a test scenario for this, but its totally understandable if that is a prerequisite for considering this PR, just let me know.

Thanks for your work on this plugin, excited to be able to pull the upstream back into my project if the team agrees this functionality is worth including in Simple Local Avatars.

t-lock:
Also, I wasn't sure where to add documentation, apologies if that was something obvious I missed. Happy to add if someone can point me in the right direction.

includes/class-simple-local-avatars.php
```diff
@@ -953,6 +953,7 @@ public function assign_new_user_avatar( $url_or_media_id, $user_id ) {
 		$meta_value['blog_id'] = get_current_blog_id();
 
 		update_user_meta( $user_id, $this->user_key, $meta_value ); // save user information (overwriting old).
+		do_action("simple_local_avatar_updated", $user_id);
```
```suggestion
		do_action( 'simple_local_avatar_updated' , $user_id );
```

includes/class-simple-local-avatars.php
```diff
@@ -1054,6 +1055,7 @@ public function action_remove_simple_local_avatar() {
 			}
 
 			$this->avatar_delete( $user_id );    // delete old images if successful
+			do_action("simple_local_avatar_deleted", $user_id);
```
```suggestion
			do_action( 'simple_local_avatar_deleted', $user_id );
```

faisal-alvi (changes requested):
A few PHPCS issues.

faisal-alvi:
@dkotter Code LGTM, any feedback from your end?

faisal-alvi (dismissed):

includes/class-simple-local-avatars.php
```diff
@@ -953,6 +953,7 @@ public function assign_new_user_avatar( $url_or_media_id, $user_id ) {
 		$meta_value['blog_id'] = get_current_blog_id();
 
 		update_user_meta( $user_id, $this->user_key, $meta_value ); // save user information (overwriting old).
+		do_action( 'simple_local_avatar_updated' , $user_id );
```
dkotter, line 964: I know we don't strictly follow this currently in this plugin but ideally any hooks we add should have documentation with them, see https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters. Here's an example in this plugin: https://github.com/10up/simple-local-avatars/blob/develop/includes/class-simple-local-avatars.php#L429.

If that could be added to this action and the other one, that would be great

dkotter (changes requested):

t-lock:
@dkotter

>If that could be added to this action and the other one, that would be great

Inline documentation added. Note: I added `@since 2.6.0` assuming that this would be included in the next release and it looks like the project isn't releasing 2.x.x minor versions. 

Thanks!

cc @faisal-alvi

dkotter (approved):

Issue #141, opened 2022-07-19
An error occurred cropping the image.


faisal-alvi:
### Describe the bug

A new bug was [reported](https://wordpress.org/support/topic/an-error-occurred-cropping-the-image/) on the support forum.

> Hi, clicking crop the image only lets me do it as admin. In roles such as author or editor I always get the error “an error occurred cropping the image”.
I have tried disabling all plug-ins, but it still gives the same error.
Any ideas, thanks.

Another user:

> I am facing exactly the same problem, any idea would be helpful, thanks!

### Steps to Reproduce

Try to crop the image as an author or editor. It will show an error and the images won't be cropped.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

faisal-alvi:
Update: Internal discussion with @ajmaurya99: https://10up.slack.com/archives/CG6QB7HEU/p1657221871321319

zamanq:
@jeffpaul Alright, did some digging. Looks like the image cropping is handled by core customize-control script which has the action "crop-image" and only users with admin capabilities can access customizer related settings. FYI: [Roles & Caps](https://wordpress.org/support/article/roles-and-capabilities/)

So in our plugin we have nothing that prevents any editor or author to crop the avatar, it's the core customizer-control which prevents users other than admins to access that.

faisal-alvi:
@zamanq do you think we should completely remove the crop feature (inherited from customizer) as this is preventing non-admins to crop images And replace it with another crop tool? (_note: we'll create a separate ticket, just asking for your opinion_)
cc: @jeffpaul

zamanq:
@faisal-alvi Having the crop feature is useful for people when setting up their avatars because it's the personalization feature that they would expect to have in their profile page. So completely removing the crop feature may not be a good idea.

Instead we could introduce a custom cropping feature using third-party library that goes well in par with WordPress and that could be a better solution.
cc: @jeffpaul

faisal-alvi:
@jeffpaul Should we publish a new release (2.6.0) this month and tackle this "new crop tool" enhancement in a separate ticket and plan it for the next release (2.7.0)?

(Optional) We can also remove the existing crop feature with 2.6.0 (with a note that it was only working for admin roles).

Please share your thoughts.

jeffpaul:
@faisal-alvi I think it's fine to keep as-is for now and can adjust for all roles with the new approach in the next release

faisal-alvi:
Closing after #150

PR #155, opened 2022-09-22
Fix: Non admin can not crop avatar

### LINKED ISSUE #150: Alternate Cropping Solution

As per discussion [here](https://github.com/10up/simple-local-avatars/issues/141#issuecomment-1224456099), we have used core's customize-control script to crop images which are only working for users with admin capabilities because only they can access customizer settings.

It was originally added in #83. It is now decided to remove it and use another crop tool that works with any cap users.

- [ ] Find a proper crop tool/script that fulfills the requirement.
- [ ] Implement and test if that works with all users.
- [ ] Confirm that users can use the 'Skip Crop' feature (already there, make it compatible with a new solution too).
- [ ] Check if that works in multisite too.
- [ ] Update the users at the [support ticket](https://wordpress.org/support/topic/an-error-occurred-cropping-the-image/) once the fix is merged and released.

### Designs

Try to keep the same as the current one.

_No response_

### Code of Conduct

jayedul:
### Description of the Change

In this pull request we've just unregistered all the `setup_theme` action hook at application runtime during ajax call only when it is a crop request. As I debugged backtrace, I found this hook execution  ([https://github.com/WordPress/WordPress/blob/master/wp-settings.php#L549](https://github.com/WordPress/WordPress/blob/master/wp-settings.php#L549)) exits somewhere in case of non admin users. 

I definitely agree it's a very important hook, but as long as we are dealing with only image crop request, so I believe here's nothing to do with theme related stuffs. That's why cleared the hook for the crop request only. 

The issue description meant to integrate another crop tool other than the core one. But that would add more dev stacks. So 
 before proceeding new stacks, this pull request just proposes a simpler fix for the original issue #141 
If it doesn't seem an appropriate solution, then no issue to adopt a new crop tool. 

Closes #150 and #141 

### How to test the Change

1. Go to user edit screen as like as you do from admin account
2. Go to avatar section and click the button `Choose From Media Library` 
3. Select Avatar and Crop
4. Crop a portion as your wish
5. Click Crop button
6. Previously this is where you faced the error, but now it should save as expected.

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -142,6 +142,13 @@ class Simple_Local_Avatars {
 				return $strings;
 			}, 10, 1 );
 		}
+
+		// Fix: An error occurred cropping the image (https://github.com/10up/simple-local-avatars/issues/141)
+		if ( isset( $_POST['action'] ) && 'crop-image' === $_POST['action'] && is_admin() && wp_doing_ajax() ) {
+			add_action( 'plugins_loaded', function() {
+				remove_all_actions( 'setup_theme' );
+			});
+		}
 	}
 
 	/**
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -143,11 +143,11 @@ class Simple_Local_Avatars {
 			}, 10, 1 );
 		}
 
-		// Fix: An error occurred cropping the image (https://github.com/10up/simple-local-avatars/issues/141)
+		// Fix: An error occurred cropping the image (https://github.com/10up/simple-local-avatars/issues/141).
 		if ( isset( $_POST['action'] ) && 'crop-image' === $_POST['action'] && is_admin() && wp_doing_ajax() ) {
-			add_action( 'plugins_loaded', function() {
+			add_action( 'plugins_loaded', function () {
 				remove_all_actions( 'setup_theme' );
-			});
+			} );
 		}
 	}
 


includes/class-simple-local-avatars.php
```diff
@@ -142,6 +142,13 @@ public function add_hooks() {
 				return $strings;
 			}, 10, 1 );
 		}
+
+		// Fix: An error occurred cropping the image (https://github.com/10up/simple-local-avatars/issues/141)
+		if ( isset( $_POST['action'] ) && 'crop-image' === $_POST['action'] && is_admin() && wp_doing_ajax() ) {
+			add_action( 'plugins_loaded', function() {
+				remove_all_actions( 'setup_theme' );
+			});
+		}
```
```suggestion
		// Fix: An error occurred cropping the image (https://github.com/10up/simple-local-avatars/issues/141).
		if ( isset( $_POST['action'] ) && 'crop-image' === $_POST['action'] && is_admin() && wp_doing_ajax() ) {
			add_action( 'plugins_loaded', function () {
				remove_all_actions( 'setup_theme' );
			} );
		}
```

faisal-alvi (changes requested):

faisal-alvi (approved):
@jayedul this is awesome! Working fine!

@dkotter Code LGTM but would like to know your opinion too before merging.

dkotter:
So if I'm following correctly, the crop tool we're currently using is also used within the customizer and it's limited there to only work for admins, correct?

If so, my concern here is that it seems with the changes here, we would also allow non-admin users the ability to crop within the customizer (though I haven't tested that). If that is the case, is there some way to fix that? For instance, could we add some unique data to our crop request that we can look for and only remove the `setup_theme` hook in that case? That way we can be sure this will only impact crop requests coming from this plugin.

faisal-alvi:
@dkotter Only users with the "customizer" cap can access the customizer area. And as per the [Roles and Capabilities](https://wordpress.org/support/article/roles-and-capabilities/) doc, the "customize" cap is only with the Super Admins and Administrators user roles. I assume this resolves your concern.

faisal-alvi:
@dkotter any thoughts on my previous comment?

dkotter:
> @dkotter any thoughts on my previous comment?

@faisal-alvi Okay, I think we're probably good to go here then. Still makes me a little wary but as long as we've tested with different account levels and haven't seen any issues with users having access they shouldn't have, should be good to go

faisal-alvi:
I've tested again and the "Appearance > Customize" is only accessible for the users with the "customizer" cap. And only admins and super admins have that cap.

faisal-alvi (approved):

PR #156, opened 2022-10-03
Address deprecation warning for wpmu_new_blog

kadamwhite:
### Description of the Change

Adapt plugin to use `wp_initialize_site` instead of deprecated `wpmu_new_blog` hook when using a recent WordPress version.

`wpmu_new_blog` has been [deprecated since 5.1.0](https://developer.wordpress.org/reference/hooks/wpmu_new_blog/), in favor of [`wp_initialize_site`](https://developer.wordpress.org/reference/hooks/wp_initialize_site/) which receives a full WP_Site instance instead of the Blog ID.

This approach of checking the version matches how [`wordpress-seo` handles this](https://github.com/Yoast/wordpress-seo/blob/9946d8cdc97857fa8311be2ab0b11cce539beb98/wp-seo-main.php#L442-L449), but depending on this plugin's support matrix, the function could alternatively be switched over to use only the new hook.

### How to test the Change

I _have not_ exhaustively tested this locally. To test, create a new network site and verify whether the `simple_local_avatars` option has the expected defaults.

### Changelog Entry

> Changed - Set plugin defaults on `wp_initialize_site` instead of deprecated action `wpmu_new_blog`.

### Checklist:

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -1239,7 +1239,7 @@ class Simple_Local_Avatars {
 	/**
 	 * Set plugin defaults for a new site
 	 *
-	 * @param int|WP_Site $blog_id Blog ID.
+	 * @param int|WP_Site $blog_id Blog ID or object.
 	 */
 	public function set_defaults( $blog_id ) {
 		if ( 'enforce' === $this->get_network_mode() ) {
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -100,6 +100,7 @@ class Simple_Local_Avatars {
 	 */
 	public function add_hooks() {
 		global $pagenow;
+		global $wp_version;
 
 		add_filter( 'plugin_action_links_' . SLA_PLUGIN_BASENAME, array( $this, 'plugin_filter_action_links' ) );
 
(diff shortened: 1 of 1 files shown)


includes/class-simple-local-avatars.php
```diff
@@ -133,7 +134,12 @@ public function add_hooks() {
 		add_action( 'wp_ajax_sla_clear_user_cache', array( $this, 'sla_clear_user_cache' ) );
 
 		add_filter( 'avatar_defaults', array( $this, 'add_avatar_default_field' ) );
-		add_action( 'wpmu_new_blog', array( $this, 'set_defaults' ) );
+		if ( version_compare( $wp_version, '5.1', '<' ) {
```
```suggestion
		if ( version_compare( $wp_version, '5.1', '<' ) ) {
```

includes/class-simple-local-avatars.php
```diff
@@ -1233,13 +1239,17 @@ public function pre_option_simple_local_avatars( $value ) {
 	/**
 	 * Set plugin defaults for a new site
 	 *
-	 * @param int $blog_id Blog ID.
+	 * @param int|WP_Site $blog_id Blog ID.
```
```suggestion
	 * @param int|WP_Site $blog_id Blog ID or object.
```

includes/class-simple-local-avatars.php
```diff
@@ -133,7 +134,12 @@ public function add_hooks() {
 		add_action( 'wp_ajax_sla_clear_user_cache', array( $this, 'sla_clear_user_cache' ) );
 
 		add_filter( 'avatar_defaults', array( $this, 'add_avatar_default_field' ) );
-		add_action( 'wpmu_new_blog', array( $this, 'set_defaults' ) );
+		if ( version_compare( $wp_version, '5.1', '<' ) {
+			add_action( 'wpmu_new_blog', array( $this, 'set_defaults' ) );
+		} else {
+			add_action( 'wp_initialize_site', array( $this, 'set_defaults' ) );
+		}
+
```
```suggestion
```

faisal-alvi (changes requested):

faisal-alvi:
@kadamwhite thanks for your work! I have made minor spacing fixes and a fatal error of a missing bracket. Then tested and it's working fine. 🎉 

The object result while testing:

image

faisal-alvi (approved):

PR #171, opened 2022-12-07
Avatar upload size filter hook fixed, file security enhanced

### LINKED ISSUE #58: Apply a proper maximum file size restriction
I'd like to be able to set a maximum upload size (100kb for example) so that my users don't upload a 10mb image.

-------------------

### Underlying problem

This has been suggested and appears that there's some code to deal with it: https://github.com/10up/simple-local-avatars/blob/e058862b7b9633ebf416b894ffac4210efa7f60e/simple-local-avatars.php#L373

But it does not work, as it bypasses the Wordpress maximum filesize checking. Even defining on the global wordpress filter

```
add_filter( 'upload_size_limit', 'my_upload_size_limit', 10, 3);
function my_upload_size_limit($i1, $i2, $i3) {
	return 100;
}
```

This last block Limits the Media upload page, but it does not restricts the Simple Local Avatar way of upload, since it calls directly `media_handle_upload`, which does not check for file size.

-------------

### Proposed solution

A proper way to solve this would be similar to what you have here https://github.com/10up/simple-local-avatars/blob/e058862b7b9633ebf416b894ffac4210efa7f60e/simple-local-avatars.php#L362 but instead doing something like this
```
if ( $_FILES['simple-local-avatar']['size'] > DEFINED_SIZE_IN_BYTES ) ) {
	//throw error here;
	return;
}
```

Of course, you could just hook the previous filter to get the information. I can submit a PR, but this is a really simple check, just need proper error throwing.

jayedul:
### Description of the Change
Avatar upload size hook added and file security enhanced.

Closes #58 

### How to test the Change

- Login to an user account who is not admin and don't have `upload_files` capability
- Upload non image file, it should show error message
- Add this filter hook anywhere to restrict upload size `add_filter('simple_local_avatars_upload_limit', function(){
	return 1024*100; // 100 KB
});` 
- Now the user should see error message if uploads images larger than the size. 

### Changelog Entry

> Added - Filter hook to restrict image upload size
> Security - Image file checking enhanced

### Credits

Props @jayedul

### Checklist:

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -1000,13 +1000,7 @@ class Simple_Local_Avatars {
 				return;
 			}
 
-			/**
-			 * Change default WP file upload limit for avatar
-			 *
-			 * @param int Max upload size in Byte
-			 */
-			$max_upload_size = apply_filters( 'simple_local_avatars_upload_limit', wp_max_upload_size() );
-
+			$max_upload_size = $this->upload_size_limit( wp_max_upload_size() );
 			if ( $_FILES['simple-local-avatar']['size'] > $max_upload_size ) {
 				$this->avatar_upload_error = sprintf( __( 'Max allowed avatar size is %s', 'simple-local-avatars' ), size_format( $max_upload_size ) );
 				add_action( 'user_profile_update_errors', array( $this, 'user_profile_update_errors' ) );
@@ -1061,6 +1055,16 @@ class Simple_Local_Avatars {
 
 			update_user_meta( $user_id, $this->rating_key, $_POST['simple_local_avatar_rating'] );
 		}
+	}
+
+	/**
+	 * Allow developers to override the maximum allowable file size for avatar uploads
+	 *
+	 * @param  int $bytes WordPress default byte size check
+	 * @return int Maximum byte size
+	 */
+	public function upload_size_limit( $bytes ) {
+		return apply_filters( 'simple_local_avatars_upload_limit', $bytes );
 	}
 
 	/**
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -988,13 +988,23 @@ class Simple_Local_Avatars {
 		}
 
 		// check for uploaded files
-		if ( ! empty( $_FILES['simple-local-avatar']['name'] ) ) :
+		if ( ! empty( $_FILES['simple-local-avatar']['name'] ) && $_FILES['simple-local-avatar']['error'] === 0 ) :
 
 			// need to be more secure since low privelege users can upload
-			if ( false !== strpos( $_FILES['simple-local-avatar']['name'], '.php' ) ) {
-				$this->avatar_upload_error = __( 'For security reasons, the extension ".php" cannot be in your file name.', 'simple-local-avatars' );
-				add_action( 'user_profile_update_errors', array( $this, 'user_profile_update_errors' ) );
+			$allowed_mime_types = wp_get_mime_types();
+			$file_mime_type = strtolower( $_FILES['simple-local-avatar']['type'] );
 
+			if ( ! ( strpos( $file_mime_type, 'image/' ) === 0 ) || ! in_array( $file_mime_type, $allowed_mime_types ) ) {
+				$this->avatar_upload_error = __( 'Only image can be uploaded as avatar', 'simple-local-avatars' );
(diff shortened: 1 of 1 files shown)


includes/class-simple-local-avatars.php
```diff
@@ -988,13 +988,23 @@ public function edit_user_profile_update( $user_id ) {
 		}
 
 		// check for uploaded files
-		if ( ! empty( $_FILES['simple-local-avatar']['name'] ) ) :
+		if ( ! empty( $_FILES['simple-local-avatar']['name'] ) && $_FILES['simple-local-avatar']['error'] === 0 ) :
```
```suggestion
		if ( ! empty( $_FILES['simple-local-avatar']['name'] ) && 0 === $_FILES['simple-local-avatar']['error'] ) :
```

includes/class-simple-local-avatars.php
```diff
@@ -988,13 +988,23 @@ public function edit_user_profile_update( $user_id ) {
...
 
 			// need to be more secure since low privelege users can upload
-			if ( false !== strpos( $_FILES['simple-local-avatar']['name'], '.php' ) ) {
-				$this->avatar_upload_error = __( 'For security reasons, the extension ".php" cannot be in your file name.', 'simple-local-avatars' );
+			$allowed_mime_types = wp_get_mime_types();
+			$file_mime_type = strtolower( $_FILES['simple-local-avatar']['type'] );
```
faisal-alvi: Needs vertical alignment.

includes/class-simple-local-avatars.php
```diff
@@ -988,13 +988,23 @@ public function edit_user_profile_update( $user_id ) {
...
+
+			if ( ! ( strpos( $file_mime_type, 'image/' ) === 0 ) || ! in_array( $file_mime_type, $allowed_mime_types ) ) {
```
```suggestion
			if ( ! ( 0 === strpos( $file_mime_type, 'image/' ) ) || ! in_array( $file_mime_type, $allowed_mime_types ) ) {
```

faisal-alvi: Please use the third argument for strict type check in`in_array`.

includes/class-simple-local-avatars.php
```diff
@@ -988,13 +988,23 @@ public function edit_user_profile_update( $user_id ) {
...
+				$this->avatar_upload_error = __( 'Only image can be uploaded as avatar', 'simple-local-avatars' );
```
```suggestion
				$this->avatar_upload_error = __( 'Only images can be uploaded as an avatar', 'simple-local-avatars' );
```

includes/class-simple-local-avatars.php
```diff
@@ -988,13 +988,23 @@ public function edit_user_profile_update( $user_id ) {
...
 				add_action( 'user_profile_update_errors', array( $this, 'user_profile_update_errors' ) );
+				return;
+			}
 
+			// Restrict file size
+			$max_upload_size = apply_filters( 'simple_local_avatar_max_size', wp_max_upload_size() );
```
faisal-alvi: We may not follow it in all places but whenever possible, we should add documentation above the `apply_filters`. Please check below example:

```
/**
 * Change default ordering support for a post type.
 *
 * @since 2.0.0
 *
 * @param boolean $sortable Whether this post type is sortable or not.
 * @param string  $post_type The post type being checked.
 */
return apply_filters( 'simple_page_ordering_is_sortable', $sortable, $post_type );
```

faisal-alvi: Also, instead of adding a new filter `simple_local_avatar_max_size`, let's use the existing filter `simple_local_avatars_upload_limit`.

faisal-alvi (changes requested):
@jayedul thanks for the PR. I have left a few comments below. Also we should remove the old code (mentioned as follows) that was supposed to handle to upload size limit.

https://github.com/10up/simple-local-avatars/blob/fb3c74de63f09fa9d3b82c84e1065e1088d820f4/includes/class-simple-local-avatars.php#L1024-L1025

https://github.com/10up/simple-local-avatars/blob/fb3c74de63f09fa9d3b82c84e1065e1088d820f4/includes/class-simple-local-avatars.php#L1043

https://github.com/10up/simple-local-avatars/blob/fb3c74de63f09fa9d3b82c84e1065e1088d820f4/includes/class-simple-local-avatars.php#L1066-L1074

And then instead of adding a new filter `simple_local_avatar_max_size`, let's use the existing filter `simple_local_avatars_upload_limit`.

faisal-alvi (approved):
@jayedul thank you for working on this. I have tested and it is working as expected! Can you please just remove the new hook mentioned in the PR description that we've removed later?

🚀

PR #180, opened 2022-12-14
Local avatar delete-remove functionality added

### LINKED ISSUE #41: Option to prevent media deletion
 We received the following from @zalemcitizen on [WordPress.ORG](https://wordpress.org/support/topic/option-to-prevent-media-deletion/):
> Hello,
> 
> I find hard that the only way to remove avatar from user is to use ‘delete local avatar’ button which deletes completely avatar image.
I need to remove avatar choice from user but keep the file in Media library because it may be used elsewhere.
> 
> Could you add a simple cross or button to break link between user and media object ?

Ability to select "delete" or "remove" local avatar where "delete" completely deletes the image and "remove" retains the image file in the media library and just disassociates it from the user.

**Designs**

**Additional context**

jayedul:
### Description of the Change
From now on, users will see a new button to disassociate avatar rather than deleting permanently since there's a possibility of using the avatar image elsewhere. 

Closes #41 

### How to test the Change

- Login to a non-admin user account that has has no `upload_files` capability
- Go to profile edit page in WP dashboard
- Set a new avatar
- Click delete, it should delete file from the file system
- Set a new avatar again
- This time just remove, it should disassociate the image as avatar, however you should find the image still in the file system

### Changelog Entry

> Added - Avatar removal feature to disassociate images only instead of deleting from file syste,

### Credits

Props @jayedul 

### Checklist:


assets/js/simple-local-avatars.js
```diff
@@ -123,21 +123,25 @@ jQuery(document).ready(function ($) {
 	/**
 	 * If the Local Avatar is removed and set to the  default one
 	 */
-	$('#simple-local-avatar-remove').on('click', function (event) {
+	$('#simple-local-avatar-remove, #simple-local-avatar-disassociate').on('click', function (event) {
 		event.preventDefault();
 
 		if (avatar_working) return;
 
+		var button_wrapper = $(this).parent();
```
faisal-alvi: Should we use `const` instead of `var`?

assets/js/simple-local-avatars.js
```diff
@@ -123,21 +123,25 @@ jQuery(document).ready(function ($) {
...
+
 		avatar_lock('lock');
 		$.get(i10n_SimpleLocalAvatars.ajaxurl, {
 			action: 'remove_simple_local_avatar',
+			permanent_delete: $(this).attr('id') === 'simple-local-avatar-remove',
```
faisal-alvi: This is confusing, can we please use the word "-delete" instead of "-remove", to match it with what we show on front side?

includes/class-simple-local-avatars.php
```diff
@@ -885,19 +885,29 @@ public function edit_user_profile( $profileuser ) {
 							// force them to use the WP Media Selector
 							// At FE, show the file input field regardless of the caps.
 							if ( ! is_admin() || ! current_user_can( 'upload_files' ) ) {
+								$old_avatars = (array) get_user_meta( $profileuser->ID, $this->user_key, true );
+								$has_local_avatar = ! empty( $old_avatars ) && ! empty( $old_avatars['media_id'] );
```
faisal-alvi: Vertical alignment is required.

includes/class-simple-local-avatars.php
```diff
@@ -885,19 +885,29 @@ public function edit_user_profile( $profileuser ) {
...
 								?>
-								<p style="display: inline-block; width: 26em;">
-									<span class="description"><?php esc_html_e( 'Choose an image from your computer:' ); ?></span><br />
-									<input type="file" name="simple-local-avatar" id="simple-local-avatar" class="standard-text" />
-									<span class="spinner" id="simple-local-avatar-spinner"></span>
-								</p>
+								<div id="simple-local-avatar-upload-interface" style="display: inline-block; width: 26em;">
+									<div style="display:<?php echo $has_local_avatar ? 'block' : 'none'; ?>">
+										Delete or remove existing avatar to set a new one.
```
faisal-alvi: Please make the string translatable.

faisal-alvi (changes requested):
@jayedul thank you for working on this.

1. I have tested and seems like when we select the image, it does not show the Remove button for `upload_files` cap users.

sla-180

2. And when the avatar is removed/deleted, we don't get back the "Choose from media library" button.

3. Can you please let me know why is the `package-lock.json` is changed?
4. Also requested a few changes inline, please have a look.

jayedul:
@faisal-alvi 
Eventually I noticed there are some inconsistent functional flow. Worked on those little bit. Now the feature should work as expected. Please let me know if still need more fix.

assets/scss/simple-local-avatars.scss
```diff
@@ -0,0 +1,13 @@
+#sla-avatar-manager{
+	&.has-local-avatar {
+		.sla-avatar-upload, .sla-avatar-choose{
```
faisal-alvi: Please add a space after `{` on line 1 and 3.

assets/js/simple-local-avatars.js
```diff
@@ -1,22 +1,25 @@
 /* eslint-disable no-undef, eqeqeq, no-use-before-define */
 
+import '../scss/simple-local-avatars.scss';
```
faisal-alvi: Why are we importing the SCSS file inside the JS file?

includes/class-simple-local-avatars.php
```diff
@@ -1125,26 +1139,28 @@ public function ajax_assign_simple_local_avatar_media() {
...
+			if ( ! empty( $old_avatars ) ) {
+				$upload_path = wp_upload_dir();
 
-			foreach ( $old_avatars as $old_avatar ) {
-				// derive the path for the file based on the upload directory
-				$old_avatar_path = str_replace( $upload_path['baseurl'], $upload_path['basedir'], $old_avatar );
-				if ( file_exists( $old_avatar_path ) ) {
-					unlink( $old_avatar_path );
+				foreach ( $old_avatars as $old_avatar ) {
+					// derive the path for the file based on the upload directory
```
```suggestion
					// Derive the path for the file based on the upload directory.
```

faisal-alvi (changes requested):
@jayedul thanks for the fixes. The remove and delete buttons are now displaying as expected. However, I have tested again and found a few more things to share, please check the following points.

1. When we hit the Delete button, it does not delete the uploaded images from the file system and we still see the image available under the Media > Library, it still keeps a few (size) versions of the image. Please check the video.

https://user-images.githubusercontent.com/25176325/212338142-ff6ef254-71a0-4f38-a941-ac3bfd46aa33.mp4

2. While testing, I felt that we should add a prompt dialog box when the user clicks the Delete button, the same as we see in the core.

image

This will be helpful to make sure that users know the difference between two buttons and that they don't get confused.

3. Please add a gap between the buttons and the text above them.
image

4. I feel that we should add more document comments (not just above this PR's code, but in the whole class file) so that we/anyone knows what's happening. This is just an opinion, you can skip this for now. :)

5. Please also check a few inline comments.

assets/js/simple-local-avatars.js
jayedul: This way we get compiled css file in `dist` directory.

faisal-alvi: We should add it as a new entry in the config so we get its compiled version in `dist`.

https://github.com/10up/simple-local-avatars/blob/65fa58eaa2d309b9b2ceec76c97b1d54be400c23/webpack.config.js#L10-L12

assets/js/simple-local-avatars.js
```diff
@@ -131,6 +129,10 @@ jQuery(document).ready(function ($) {
 
 		if (avatar_working) return;
 
+		if($(this).data('alert') && !window.confirm($(this).data('alert'))) {
```
faisal-alvi:
```suggestion
		if ($(this).data('alert') && !window.confirm($(this).data('alert'))) {
```

includes/class-simple-local-avatars.php
```diff
@@ -935,7 +935,7 @@ public function edit_user_profile( $profileuser ) {
 								<div>
 									<?php esc_html_e( 'Delete or remove existing avatar to set a new one.', 'simple-local-avatars' ); ?> 
 								</div>
-								<a href="<?php echo esc_url( $remove_url ); ?>" class="button item-delete submitdelete deletion" id="simple-local-avatar-delete">
+								<a href="<?php echo esc_url( $remove_url ); ?>" class="button item-delete submitdelete deletion" id="simple-local-avatar-delete" data-alert="<?php echo esc_attr__( 'Are you sure to delete the image permanently?', 'simple-local-avatars' ); ?>">
```
```suggestion
								<a href="<?php echo esc_url( $remove_url ); ?>" class="button item-delete submitdelete deletion" id="simple-local-avatar-delete" data-alert="<?php echo esc_attr__( 'Are you sure that you want to permanently delete the image?', 'simple-local-avatars' ); ?>">
```

faisal-alvi (changes requested):

faisal-alvi (approved):
@jayedul thanks for the work, LGTM!

faisal-alvi:
@jayedul During testing, I observed that when someone permanently deletes an image, if anyone else is using the same avatar, their avatar will switch to the default avatar, which may lead to confusion among users. Just wanted to bring it to everone's attention. cc: @dkotter

dkotter:
@faisal-alvi @jayedul If I'm following correctly, the `Delete` button is an existing feature within this plugin but it wasn't actually deleting the image fully, is that correct? And this PR fixes that as well as adds a `Remove` button?

jayedul:
@dkotter Exactly. 

@faisal-alvi To avoid confusion, I think it will be better show only own uploaded images to select avatar from. So, no way to use same image by multiple user. Then deletion will be safe.

dkotter:
@jayedul @faisal-alvi @jeffpaul Apologies for not flagging things sooner but the more I think through things here, the more I think we may have gone further then what is actually needed here.

Reading through the [issue](https://github.com/10up/simple-local-avatars/issues/41) again, the problem being reported is they want a way to remove an avatar image from a specific user without having to delete that image all together. But that functionality is already in place.

I think the confusion here is the button we output says `Delete local avatar` but it doesn't actually delete the image from what I can tell, it just removes that image from being the user's avatar.

I could see having actual delete functionality to be useful in some scenarios but as @jayedul as already mentioned, you could end up in a situation where someone deletes an image that is used elsewhere, unless we lock down images to the specific user.

For a plugin that we consider stable, I guess my opinion here is that we don't need to add true deletion functionality at the moment. I'd suggest either leaving things as they are or updating the button text to say `Remove local avatar`, so it's more clear what the button does.

faisal-alvi:
@dkotter, thanks for the feedback. I agree with you. For now, we can update the button text to say `Remove local avatar`. We can use the code from this PR in the future if required to add permanent delete functionality.

faisal-alvi:
Closing this in favor of #208

PR #183, opened 2023-01-09
Fix TypeError in PHP8

### LINKED ISSUE #179: 'assign_new_user_avatar' exception when passing url as $url_or_media_id with php 8
### Describe the bug

When the `assign_new_user_avatar` function is called with first parameter (`$url_or_media_id`) being an url string and not a numeric media id, the function throws and exception on line 959 when running under PHP 8 :

```
Uncaught TypeError: Unsupported operand types: string + int in /SomePath/simple-local-avatars/includes/class-simple-local-avatars.php:959
Stack trace:
#0 /SomePath/SimpleLocalAvatarPlugin.php(56): Simple_Local_Avatars->assign_new_user_avatar('...', 56)
...
```

### Steps to Reproduce

1. Call programmatically the `assign_new_user_avatar` like this :

```
global $simple_local_avatars;

if ( isset( $simple_local_avatars ) ) {
	$current_user = wp_get_current_user();
	$simple_local_avatars->assign_new_user_avatar( 'some_url_string', $current_user->ID );
}
```

_No response_

### Environment information

_No response_

### WordPress information

Wordpress 6.1.1
PHP 8.0.15
SimpleLocalAvatars  2.7.1

### Code of Conduct

mattwatsoncodes:
### Description of the Change

Current implementation causes an error in PHP8:

```
TypeError: Unsupported operand types: string + int
```

Using `is_numeric` instead of the additional of a 0 and `is_int` fixes the error.

Closes #179 

### How to test the Change

- Ensure the server is running PHP8
- Call the method `assign_new_user_avatar` from the `Simple_Local_Avatars` class (see https://github.com/10up/simple-local-avatars/blob/develop/includes/class-simple-local-avatars.php#L958)
- If all is working correctly a fatal error should not be triggered.

### Changelog Entry

> Added - New feature
> Changed - Existing functionality
> Deprecated - Soon-to-be removed feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

Fixed - Added PHP8 support for assign_new_user_avatar

### Credits

Props @username, @username2, ...

@mattwatsoncodes

### Checklist:

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -955,7 +955,7 @@ class Simple_Local_Avatars {
 		$meta_value = array();
 
 		// set the new avatar
-		if ( is_int( $url_or_media_id + 0 ) ) {
+		if ( is_numeric( $url_or_media_id ) ) {
 			$meta_value['media_id'] = $url_or_media_id;
 			$url_or_media_id        = wp_get_attachment_url( $url_or_media_id );
 		}


mattwatsoncodes:
Just adding a comment here around the use of `is_numeric`.

Reading through the history of the ticket a solution was originally proposed using `is_numeric` but a change was requested because `is_numeric` will return true for a float.

This issue here is `is_int` and `is_float` won't return true if the value is a string, so we cannot use these as a replacement for `is_numeric` in this scenario.

The input should always either be a string (URL of file path) or an int (which could be a  number in a string) (ID of attachment). In theory we could check if the string is numeric, if it is parse it to a float, and then check if it is a whole number, but that seemed a little overkill.

Happy to adjust the PR to have these checks in place if required though.

faisal-alvi (approved):
@mattwatsoncodes thank you for working on this. LGTM, working as expected.

mikelittle (commented):
Perhaps:
```
if ( (int) $url_or_media > 0 ) {
...
``` 
is a little more straightforward.

Issue #157, opened 2022-10-07
The plugin does not respect the user profile language setting in the backend


lllopo:
### Describe the bug

The plugin does not respect the user profile language setting in the backend.

### Steps to Reproduce

1. Set the site language to language A
2. Set the, let's say Administrator language in his profile to language B
3. Login as the Administrator in the backend and go to plugin settings.

What happens: the plugin settings are displayed in language A
Expected behavior: the plugin settings should be displayed in language B

### Environment information

Doesn't matter

### WordPress information

WordPress 6.0.2
Simple Local Avatars 2.6.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

lllopo:
Same goes for the added settings in the "Discussion" section.

jeffpaul:
@lllopo is there a translation on translate.wordpress.org available for language B?

lllopo:
@jeffpaul Most certainly yes, since my profile language (or so-called language B) is the default Wordpress English. The site language is Dutch and unfortunately the whole admin shows in English (as per my profile preferences) bar the Simple Local Avatar settings which show in Dutch (also one more plugin that has the same problem, but that's irrelevant).

lllopo:
@jeffpaul I'm the site technical administrator, but I don't speak Dutch. As you can imagine - it is not just a cosmetical annoyance, it is quite a problem for me to do configurations while settings show up in Dutch.

faisal-alvi:
I have tested and I was able to regenerate this issue, it's a weird behavior - as IMO, it should be handled by the WordPress core itself, or perhaps I might be missing something.

dkotter:
@faisal-alvi All right, I believe I have this one figured out. In doing some step through debugging, was seeing that our translations are being loaded prior to the function `wp_get_current_user` being available. This means we always end up using the site language translation files instead of the individual user's language.

This stirred a memory and I remembered we had a similar issue reported on .org that I added as an issue in #88. This was solved in #89 but since we still had strings being translated early, it seems we still have issues as reported here.

I _think_ the easiest fix here is to remove the textdomain from those [avatar rating fields](https://github.com/10up/simple-local-avatars/blob/develop/includes/class-simple-local-avatars.php#L71), as these strings are the exact same as what core uses (I believe) and as such, there's no need for us to maintain those translations.

In testing locally, I changed:
```php
$this->avatar_ratings = array(
[...]

PR #214, opened 2023-05-17
Improve I18N Issues (Based on 2.7.5)

alexclassroom:
### Description of the Change
- If there are strings in your plugin that are also used in WordPress core (e.g. **‘Settings’**), you should still add your own text domain to them, otherwise, they’ll become untranslated if the core string changes (which happens). Please refer to this [official article](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#add-text-domain-to-strings).
- Correct a UI string with the wrong text domain.

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -526,13 +526,13 @@ class Simple_Local_Avatars {
 		 */
 		$this->avatar_ratings = array(
 			/* translators: Content suitability rating: https://en.wikipedia.org/wiki/Motion_Picture_Association_of_America_film_rating_system */
-			'G'  => __( 'G &#8212; Suitable for all audiences' ),
+			'G'  => __( 'G &#8212; Suitable for all audiences', 'simple-local-avatars' ),
 			/* translators: Content suitability rating: https://en.wikipedia.org/wiki/Motion_Picture_Association_of_America_film_rating_system */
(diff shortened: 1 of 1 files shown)


faisal-alvi:
Hi @alexclassroom thank you for the report and the PR.

We have already discussed this point. There were issues reported in #88 and #157 about the user profile language setting not being respected and hence we applied a fix in #175 which removes the text domain from the core strings. Read more details by @dkotter at https://github.com/10up/simple-local-avatars/issues/157#issuecomment-1320333432.

dkotter:
So I'm not opposed to adding textdomains here and I understand that is typically best practice. I believe my original point still stands though, that if we add textdomains for core strings, we are now responsible for providing translations for those strings instead of being able to rely on core for that.

In this case, we have copied these strings straight from core and so by leaving the textdomain off, we can take advantage of the translations that are already in core. But yes, if those strings ever change, those translations will no longer work. Ideally though, if those strings change in core, we should be changing those here in Simple Local Avatars so we maintain consistency.

So while yes, it is recommended best practice to add textdomains to all strings, since these strings are copied from core and our goal is to always have these strings match what core uses, I'd prefer to leave the textdomain off and rely on translations loading from core. If we add the textdomain to these, those strings will no longer be translated until our language files get updated (which may or may not ever happen, depending on the language and community interest in adding those).

All that said, I'm not too strongly opinionated here so happy to change my thoughts on this if there's disagreement to this approach.

faisal-alvi:
Closing for now.

PR #237, opened 2023-09-21
Introduce a filter that short-circuits the avatar URL generation

johnbillion:
### Description of the Change

When a dynamic image resizing service (such as Tachyon or Photon) is in use, Simple Local Avatars provides no way for the URL of an avatar of a given size to be short-circuited so as to avoid its own dynamic resizing that happens via a `WP_Image_Editor` instance.

This change introduces a filter to short-circuit the URL for an avatar (once it's been determined that a "full" size exists), allowing a dynamic image resizing service to return its own scaled version of the user's image.

### How to test the Change

Here's a simplified example of how this filter could be used on a site that uses Tachyon:

```php
add_filter(
	'pre_simple_local_avatar_url',
	function( $url, $user_id, $size, $local_avatars ) {
		return tachyon_url( $local_avatars['full'], [ 'resize' => "{$size},{$size}" ] );
	},
	10,
	4
);
```

### Changelog Entry

> Added - `pre_simple_local_avatar_url` filter to allow an avatar image to be short-circuited before Simple Local Avatars processes it.

### Credits

@johnbillion

### Checklist:


jeffpaul:
Thanks @johnbillion, that usage makes sense to me.  Tagging @faisal-alvi for code review/testing to see if there are any concerns here before moving towards merge/release.

peterwilsoncc:
@faisal-alvi I have a copy of Tachyon set up locally if you need guidance on getting it going/want me to test.

faisal-alvi:
@peterwilsoncc it would be great if you test + code review this as I do not have Tachyon set up locally. Thanks! ❤️

includes/class-simple-local-avatars.php
```diff
@@ -353,6 +353,27 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
...
+		 * @param int         $user_id       The user ID.
+		 * @param int         $size          Requested avatar size.
+		 * @param array       $local_avatars The local avatars for the user.
+		 * @return string|null The URL of the avatar, or null to allow Simple Local Avatars to
+		 *                     generate the URL.
+		 */
+		$url = apply_filters( 'pre_simple_local_avatar_url', null, $user_id, $size, $local_avatars );
+
+		if ( is_string( $url ) ) {
+			return $url;
```
peterwilsoncc: The function is considered "safe" -- WP handles double escaping.

But hold off on actioning this given the comment above.

```suggestion
			return esc_url( $url );
```

peterwilsoncc (commented):
This tests well and works as described.

I've just raised a discussion internally as to whether the plugin should switch to using `wp_get_attachment_image_src()[0]` and storing dynamically resized images in the image post meta. That will allow the plugin to just work with Tachyon, Photon and other services using the `image_downsize` filter to dynamically resize images.

includes/class-simple-local-avatars.php
johnbillion: I think that would be to be esc_url_raw anyway

peterwilsoncc (commented):
@johnbillion We've considered the options and decided to put this in for now rather than switch to the proper media functions. The team needs to do a little more investigating of the implications for multisite installations using shared avatars. 

> I think that would be to be esc_url_raw anyway

Yeah, it probably should be but we use `esc_url()` elsewhere in the function so it's intended for printing rather than getting.. It's probably too late to make the change now.

peterwilsoncc (approved):

PR #261, opened 2024-03-10
Fix - #224 Avoid php warnings

### LINKED ISSUE #224: PHP warnings due to wrong param type
### Describe the bug

As reported on [WordPress.org](https://wordpress.org/support/topic/error-during-authorization-to-rocketchat/), seems there's some situations that can result in PHP warnings. I have not been able to reproduce myself but in looking at the code, I think we can add a couple checks here to easily suppress these warnings.

### Steps to Reproduce

See https://wordpress.org/support/topic/error-during-authorization-to-rocketchat/

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

BhargavBhandari90:
### Description of the Change
Add checks to avoid PHP warnings.

Closes #224

### How to test the Change
- I was not able to replicate the warning. But it is good to add checks, especially for the array for php8.

### Changelog Entry

> Fixed - Possible PHP warning

### Credits
Props @dkotter

### Checklist:

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -347,7 +347,7 @@ class Simple_Local_Avatars {
 		$all_avatar_ratings = ! empty( $this->avatar_ratings ) && is_array( $this->avatar_ratings )
 			? $this->avatar_ratings
 			: array();
-		if ( ! empty( $avatar_rating ) && 'G' !== $avatar_rating && $site_rating && $all_avatar_ratings ) {
+		if ( ! empty( $avatar_rating ) && 'G' !== $avatar_rating && $site_rating ) {
 			$ratings              = array_keys( $all_avatar_ratings );
 			$site_rating_weight   = array_search( $site_rating, $ratings, true );
 			$avatar_rating_weight = array_search( $avatar_rating, $ratings, true );
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -342,10 +342,13 @@ class Simple_Local_Avatars {
 		}
 
 		// check rating
-		$avatar_rating = get_user_meta( $user_id, $this->rating_key, true );
-		$site_rating   = get_option( 'avatar_rating' );
-		if ( ! empty( $avatar_rating ) && 'G' !== $avatar_rating && $site_rating ) {
-			$ratings              = array_keys( $this->avatar_ratings );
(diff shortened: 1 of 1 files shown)


includes/class-simple-local-avatars.php
```diff
@@ -342,10 +342,13 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
...
-		$avatar_rating = get_user_meta( $user_id, $this->rating_key, true );
-		$site_rating   = get_option( 'avatar_rating' );
-		if ( ! empty( $avatar_rating ) && 'G' !== $avatar_rating && $site_rating ) {
-			$ratings              = array_keys( $this->avatar_ratings );
+		$avatar_rating      = get_user_meta( $user_id, $this->rating_key, true );
+		$site_rating        = get_option( 'avatar_rating' );
+		$all_avatar_ratings = ! empty( $this->avatar_ratings ) && is_array( $this->avatar_ratings )
+			? $this->avatar_ratings
+			: array();
+		if ( ! empty( $avatar_rating ) && 'G' !== $avatar_rating && $site_rating && $all_avatar_ratings ) {
```
dkotter: Do we need the `$all_avatar_ratings` check here? Unless I'm missing something, in looking at the code changes here, `$all_avatar_ratings` will either be an array of ratings or will be an empty array which in either case, this check will always return `true` I think.

BhargavBhandari90: Yes, even I was not getting the warning reported by the user mentioned at this [issue](https://wordpress.org/support/topic/error-during-authorization-to-rocketchat/)

I've just added a check for the safer side just in case some other plugin conflicts with this plugin.

dkotter: I guess to be more clear, I do think this code is great:
```php
$all_avatar_ratings = ! empty( $this->avatar_ratings ) && is_array( $this->avatar_ratings )
	? $this->avatar_ratings
	: array();
```

My question is on the addition of `$all_avatar_ratings` in the `if` statement. It doesn't seem like that will ever do anything (that value will always be an `array` now and thus having it in that statement will never change things). Unless I'm missing something, I think we can remove that piece:
```suggestion
		if ( ! empty( $avatar_rating ) && 'G' !== $avatar_rating && $site_rating ) {
```

BhargavBhandari90: Oh yes. Got it. Sorry, I misunderstood your query. Yes, we can remove that from `if`.

dkotter (approved):

PR #302, opened 2024-08-28
fix: user meta retrieval

adekbadek:
### Description of the Change

Add sanity checks to the user meta values read from the DB. We've seen this user data malformatted – maybe it was in migration from another WP instance. 

### How to test the Change

1. Directly in the DB, update the `simple_local_avatar` meta of a user to an empty string
2. Update the `simple_local_avatar` meta of another user to an object (e.g. `update_user_meta(<id>, 'simple_local_avatar', (object) ['value' => 42])`)
3. Observe that these changes crash the site on `develop` branch 
4. Switch to this branch, observe the site is not crashing

### Changelog Entry

> Fixed - Handling of malformed `simple_local_avatar` user meta.

### Checklist:

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -358,7 +358,7 @@ class Simple_Local_Avatars {
 
 		// Fetch local avatar from meta and make sure it's properly set.
 		$local_avatars = $this->get_user_local_avatar( $user_id );
-		if ( empty( $local_avatars['full'] ) ) {
+		if ( ! isset( $local_avatars['full'] ) ||empty( $local_avatars['full'] ) ) {
 			return '';
 		}
 
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -332,7 +332,7 @@ class Simple_Local_Avatars {
 	 * @param int $user_id User ID.
 	 * @return array Array with avatar data.
 	 */
-	public static function get_user_local_avatar( $user ) {
+	public function get_user_local_avatar( $user_id ) {
 		$local_avatars = get_user_meta( $user_id, $this->user_key, true );
 		if ( ! is_array( $local_avatars ) || empty( $local_avatars ) ) {
 			return [];
@@ -357,7 +357,7 @@ class Simple_Local_Avatars {
 		}
 
 		// Fetch local avatar from meta and make sure it's properly set.
-		$local_avatars = self::get_user_local_avatar( $user_id );
+		$local_avatars = $this->get_user_local_avatar( $user_id );
 		if ( empty( $local_avatars['full'] ) ) {
 			return '';
 		}
@@ -493,7 +493,7 @@ class Simple_Local_Avatars {
 		}
 
 		// Fetch local avatar from meta and make sure we have a media ID.
-		$local_avatars = get_user_meta( $user_id, 'simple_local_avatar', true );
+		$local_avatars = $this->get_user_local_avatar( $user_id );
 		if ( empty( $local_avatars['media_id'] ) ) {
 			$alt = '';
 			// If no avatar is set, check if we are using a default avatar with alt text.
@@ -1125,7 +1125,7 @@ class Simple_Local_Avatars {
 		endif;
 
 		// Handle ratings
-		if ( isset( $avatar_id ) || get_user_meta( $user_id, $this->user_key, true ) ) {
+		if ( isset( $avatar_id ) || ! empty( $this->get_user_local_avatar( $user_id ) ) ) {
 			if ( empty( $_POST['simple_local_avatar_rating'] ) || ! array_key_exists( $_POST['simple_local_avatar_rating'], $this->avatar_ratings ) ) {
 				$_POST['simple_local_avatar_rating'] = key( $this->avatar_ratings );
 			}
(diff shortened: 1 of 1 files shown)


includes/class-simple-local-avatars.php
```diff
@@ -326,6 +326,20 @@ public function get_user_id( $id_or_email ) {
 		return $user_id;
 	}
 
+	/**
+	 * Get the local avatar user meta.
+	 *
+	 * @param int $user_id User ID.
+	 * @return array Array with avatar data.
+	 */
+	public static function get_user_local_avatar( $user ) {
```
dkotter: We use the variable `$user_id` but the only variable passed in is `$user`. I think that just needs updated:

```suggestion
	public static function get_user_local_avatar( $user_id ) {
```

dkotter: Because this method is marked as `static`, we get an error when trying to use `$this` later on

includes/class-simple-local-avatars.php
```diff
@@ -343,7 +357,7 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
 		}
 
 		// Fetch local avatar from meta and make sure it's properly set.
-		$local_avatars = get_user_meta( $user_id, $this->user_key, true );
+		$local_avatars = self::get_user_local_avatar( $user_id );
```
dkotter: If we change the above method to not be static, this will need updated:
```suggestion
		$local_avatars = $this->get_user_local_avatar( $user_id );
```

includes/class-simple-local-avatars.php
```diff
@@ -1181,7 +1195,7 @@ public function ajax_assign_simple_local_avatar_media() {
 	 * @param int $user_id User ID.
 	 */
 	public function avatar_delete( $user_id ) {
-		$old_avatars = (array) get_user_meta( $user_id, $this->user_key, true );
+		$old_avatars = self::get_user_local_avatar( $user_id );
```
dkotter:
```suggestion
		$old_avatars = $this->get_user_local_avatar( $user_id );
```

dkotter (changes requested):
In addition to the requested changes, I see a few other places that we are doing a `get_user_meta` call that probably make sense to update to this new `get_user_local_avatar` method:

- https://github.com/10up/simple-local-avatars/blob/develop/includes/class-simple-local-avatars.php#L1114
- https://github.com/10up/simple-local-avatars/blob/develop/includes/class-simple-local-avatars.php#L1275

adekbadek:
Hi @dkotter , thanks for the review! I've added some changes in d9ff4bb302c3a36d70443aefc5c1273ff2187bb3

includes/class-simple-local-avatars.php
```diff
@@ -343,7 +357,7 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
 		}
 
 		// Fetch local avatar from meta and make sure it's properly set.
-		$local_avatars = get_user_meta( $user_id, $this->user_key, true );
+		$local_avatars = $this->get_user_local_avatar( $user_id );
 		if ( empty( $local_avatars['full'] ) ) {
```
faisal-alvi: This condition might cause a PHP notice or warning if the `get_user_local_avatar()` method returns an array that doesn't have the `full` key or if an empty array is returned. To avoid the notice, we can check if the key exists before accessing it.

faisal-alvi (changes requested):

jeffpaul:
@adekbadek thanks for the PR; are you able to respond to the code review items here?

includes/class-simple-local-avatars.php
adekbadek: Added a check in eeb66bcdf86d6ca7d455edd2f363368603a97911

adekbadek:
@jeffpaul – sorry for the late reply – I've addressed the review now.

includes/class-simple-local-avatars.php
```diff
@@ -343,8 +357,8 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
 		}
 
 		// Fetch local avatar from meta and make sure it's properly set.
-		$local_avatars = get_user_meta( $user_id, $this->user_key, true );
-		if ( empty( $local_avatars['full'] ) ) {
+		$local_avatars = $this->get_user_local_avatar( $user_id );
+		if ( ! isset( $local_avatars['full'] ) ||empty( $local_avatars['full'] ) ) {
```
```suggestion
		if ( ! isset( $local_avatars['full'] ) || empty( $local_avatars['full'] ) ) {
```

includes/class-simple-local-avatars.php
```diff
@@ -479,7 +493,7 @@ public function get_simple_local_avatar_alt( $id_or_email ) {
 		}
 
 		// Fetch local avatar from meta and make sure we have a media ID.
-		$local_avatars = get_user_meta( $user_id, 'simple_local_avatar', true );
+		$local_avatars = $this->get_user_local_avatar( $user_id );
 		if ( empty( $local_avatars['media_id'] ) ) {
```
```suggestion
		if ( ! isset( $local_avatars['media_id'] ) || empty( $local_avatars['media_id'] ) ) {
```

faisal-alvi (changes requested):

faisal-alvi (approved):

PR #319, opened 2024-11-14
Update outdated dependencies

dkotter:
### Description of the Change

This PR updates all of our outdated direct node dependencies.

It also resolves two security issues with dependencies other dependencies are including. Dependabot normally fixes those but it seems to be having issues with these two (for those with access, can see this under the `Security > Dependabot` setting).

Here's what I did:

1. For any dependencies we are directly managing in our `package.json` file, bump those to the latest version, except for `mochawesome-json-to-md` as this is known to break our E2E workflow
2. Remove the existing `package-lock.json` file
3. Run `npm install` to bring in all the latest versions of packages
4. Run `npm audit` to see if any security issues remain
5. Two issues were flagged, one with `@sentry/node` and one with `puppeteer-core`. These are dependencies used by the `lighthouse` dependency, which is relied on by `@wordpress/scripts`. We're already using the latest version of `@wordpress/scripts` so updating that won't fix things. Instead, I directly updated those dependencies to their latest in the `package-lock.json` file. This fixes the security issues for now though those updates will be wiped out the next time we update `@wordpress/scripts`

### How to test the Change

Verify all tests pass on this PR

### Changelog Entry

> Security - Bump `@10up/cypress-wp-utils` from 0.2.0 to 0.4.0
> Security - Bump `@sentry/node` from 6.19.7 to 8.38.0
> Security - Bump `@wordpress/env` from 9.2.0 to 10.11.0
> Security - Bump `cypress` from 13.2.0 to 13.15.2
> Security - Bump `cypress-mochawesome-reporter` from 3.6.0 to 3.8.2
> Security - Bump `puppeteer-core` from 23.3.0 to 23.8.0

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,32 +9,32 @@
       "version": "2.8.2",
       "license": "GPL-2.0-or-later",
       "devDependencies": {
-        "@10up/cypress-wp-utils": "^0.2.0",
-        "@wordpress/env": "^9.2.0",
+        "@10up/cypress-wp-utils": "^0.4.0",
+        "@wordpress/env": "^10.11.0",
         "@wordpress/scripts": "^30.4.0",
-        "cypress": "^13.2.0",
+        "cypress": "^13.15.2",
         "cypress-file-upload": "^5.0.8",
-        "cypress-mochawesome-reporter": "^3.6.0",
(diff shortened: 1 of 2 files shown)


jeffpaul:
Great description, thanks for getting these updated!

PR #324, opened 2025-01-07
Don't create duplicate image sizes

### LINKED ISSUE #323: Don't resize image if the full version already has the expected height/width

https://github.com/10up/simple-local-avatars/blob/c5f0981a239672889913d8b525b4fc8043abd8ee/includes/class-simple-local-avatars.php#L437-L456

If the image is 160x160 and an avatar for 160x160px is requested another image with the suffix -160x160 is generated. I think we can just use the full image as set in https://github.com/10up/simple-local-avatars/blob/c5f0981a239672889913d8b525b4fc8043abd8ee/includes/class-simple-local-avatars.php#L425

We can use `$editor->get_size()` to compare the size of image with the requested size.

### Designs

_No response_

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ocean90:
### Description of the Change
Check size of original image before creating a duplicate version of it.

Closes #323

### How to test the Change
Upload an image of an avatar size that is used on the site. Ensure that no version is created for this size.

### Credits

Props @ocean90 

### Checklist:

### DIFF
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -437,20 +437,23 @@ class Simple_Local_Avatars {
 				// generate the new size
 				$editor = wp_get_image_editor( $avatar_full_path );
 				if ( ! is_wp_error( $editor ) ) {
-					$resized = $editor->resize( $size, $size, true );
-					if ( ! is_wp_error( $resized ) ) {
-						$dest_file = $editor->generate_filename();
-						$saved     = $editor->save( $dest_file );
-						if ( ! is_wp_error( $saved ) ) {
-							// Transform the destination file path into URL.
-							$dest_file_url = '';
-							if ( false !== strpos( $dest_file, $upload_path['basedir'] ) ) {
-								$dest_file_url = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
-							} elseif ( is_multisite() && false !== strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
-								$dest_file_url = str_replace( ABSPATH . 'wp-content/uploads', network_site_url( '/wp-content/uploads' ), $dest_file );
-							}
+					$image_size = $editor->get_size();
+					if ( $image_size['width'] !== $size || $image_size['height'] !== $size ) {
+						$resized = $editor->resize( $size, $size, true );
+						if ( ! is_wp_error( $resized ) ) {
+							$dest_file = $editor->generate_filename();
+							$saved     = $editor->save( $dest_file );
+							if ( ! is_wp_error( $saved ) ) {
+								// Transform the destination file path into URL.
+								$dest_file_url = '';
+								if ( false !== strpos( $dest_file, $upload_path['basedir'] ) ) {
+									$dest_file_url = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
+								} elseif ( is_multisite() && false !== strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
+									$dest_file_url = str_replace( ABSPATH . 'wp-content/uploads', network_site_url( '/wp-content/uploads' ), $dest_file );
+								}
 
-							$local_avatars[ $size ] = $dest_file_url;
+								$local_avatars[ $size ] = $dest_file_url;
+							}
 						}
 					}
 				}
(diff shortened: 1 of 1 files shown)


includes/class-simple-local-avatars.php
```diff
@@ -437,20 +437,23 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
...
-						if ( ! is_wp_error( $saved ) ) {
-							// Transform the destination file path into URL.
-							$dest_file_url = '';
-							if ( false !== strpos( $dest_file, $upload_path['basedir'] ) ) {
-								$dest_file_url = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
-							} elseif ( is_multisite() && false !== strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
-								$dest_file_url = str_replace( ABSPATH . 'wp-content/uploads', network_site_url( '/wp-content/uploads' ), $dest_file );
-							}
+					$image_size = $editor->get_size();
+					if ( $image_size['width'] !== $size || $image_size['height'] !== $size ) {
```
faisal-alvi: The `get_size()` might return `false` in some cases, so it’s good to check before accessing array keys.

faisal-alvi (changes requested):

includes/class-simple-local-avatars.php
```diff
@@ -437,20 +437,23 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
...
-						if ( ! is_wp_error( $saved ) ) {
-							// Transform the destination file path into URL.
-							$dest_file_url = '';
-							if ( false !== strpos( $dest_file, $upload_path['basedir'] ) ) {
-								$dest_file_url = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
-							} elseif ( is_multisite() && false !== strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
-								$dest_file_url = str_replace( ABSPATH . 'wp-content/uploads', network_site_url( '/wp-content/uploads' ), $dest_file );
-							}
+					$image_size = $editor->get_size();
+					if ( ! $image_size || $image_size['width'] !== $size || $image_size['height'] !== $size ) {
```
faisal-alvi: Let's ensure `$image_size` is actually an array before accessing keys
```suggestion
					if ( ! is_array( $image_size ) || $image_size['width'] !== $size || $image_size['height'] !== $size ) {
```

faisal-alvi (changes requested):
@ocean90 Thanks for the fix! Could you take a look at why the PHPUnit and Cypress tests started failing? Also, I’ve added a couple of minor suggestions below.

includes/class-simple-local-avatars.php
```diff
@@ -437,20 +437,23 @@ public function get_simple_local_avatar_url( $id_or_email, $size ) {
...
-								$dest_file_url = str_replace( $upload_path['basedir'], $upload_path['baseurl'], $dest_file );
-							} elseif ( is_multisite() && false !== strpos( $dest_file, ABSPATH . 'wp-content/uploads' ) ) {
-								$dest_file_url = str_replace( ABSPATH . 'wp-content/uploads', network_site_url( '/wp-content/uploads' ), $dest_file );
-							}
+					$image_size = $editor->get_size();
+					if ( ! is_array( $image_size ) || $image_size['width'] !== $size || $image_size['height'] !== $size ) {
+						$resized = $editor->resize( $size, $size, true );
+						if ( ! is_wp_error( $resized ) ) {
+							$dest_file = $editor->generate_filename();
+							$saved     = $editor->save( $dest_file );
```
faisal-alvi, line 445: I know this isn't new code, but do you think we should account for the case where `generate_filename()` returns `false`?

faisal-alvi (approved):

includes/class-simple-local-avatars.php
jeffpaul: @ocean90 any thoughts on the question above?

ocean90: Please feel free to change this if you think it's needed.

jeffpaul: @faisal-alvi if you feel that should be updated, mind pushing that here?

faisal-alvi: Okay, I was mistaken earlier—I'm not sure where I saw the core function returning `false`, but after reviewing the code, no changes are needed here regarding checking for a `false` return from [generate_filename()](https://developer.wordpress.org/reference/classes/wp_image_editor/generate_filename/). The current implementation is correct. The only error handling required is already present for `$editor->save()`, which can return a `WP_Error`.

faisal-alvi (approved):
@ocean90 Nice work!

PR #337, opened 2025-07-07
Fix/workflow perms

jeffpaul:
### Description of the Change

This pull request updates multiple GitHub Actions workflows to include explicit `permissions` configurations for improved security and compliance. Additionally, it simplifies the PHPUnit workflow by removing unnecessary steps.

### Permissions configuration updates:

* [`.github/workflows/build-release-zip.yml`](diffhunk://#diff-948bce84fd4ffebcbf8f82de6162123178817db71cecfb56a7fa79a9535676ecR1-R2): Added `permissions` with `contents: read`.
* [`.github/workflows/cypress.yml`](diffhunk://#diff-5da9a41590fb261de0e464c95862528100c42c77dffad53b7c6de228b6afd2bdR7-R9): Added `permissions` with `contents: read` and `actions: write`.
* [`.github/workflows/php-compatibility.yml`](diffhunk://#diff-93c2990b81df932a8ebe1a2f9ac5167d82d3193205e8878d0987d10048061e3fR3-R5): Added `permissions` with `contents: read`.
* [`.github/workflows/push-asset-readme-update.yml`](diffhunk://#diff-da3ba00836cc30a1497ac5cd5b51eba2baa577ae0a53e64a2263cc6f849e444bR8-R10): Added `permissions` with `contents: read`.
* [`.github/workflows/push-deploy.yml`](diffhunk://#diff-d71c3ba30cee07edf9485b33d2e7ee90f8abcb8c853798bea3437187ebb055bbR11-R14): Added `permissions` with `contents: read`, `packages: read`, and `id-token: write`.
* [`.github/workflows/repo-automator.yml`](diffhunk://#diff-1c38000792e37fe51624d4be0089820da8c58940b859fbbc85561037d1a07118R1-R3): Added `permissions` with `contents: read` and `pull-requests: write`.
* [`.github/workflows/test.yml`](diffhunk://#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R1-R2): Added `permissions` with `contents: read`.

### PHPUnit workflow simplification:

* [`.github/workflows/test.yml`](diffhunk://#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88L28-R30): Removed steps for dependency updates, PHPUnit tests, and network tests. Simplified `Set PHP version` step by removing additional parameters.

### Changelog Entry

> Developer - Updated GitHub Action workflow permissions.

### Credits

Props @jeffpaul.

[...]

PR #340, opened 2025-07-14
Ensure the release asset can be attached properly

dkotter:
### Description of the Change

After the last release (2.8.4), an error occured when our GitHub Action tried to attach the final release asset to the release (see https://github.com/10up/simple-local-avatars/actions/runs/16271766801). This is due to permission changes we made in #337. This PR updates the permissions there to ensure this works properly.

### How to test the Change

Nothing really to test here

### Changelog Entry

> Developer - Ensure our final release asset gets attached properly to the release

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/.github/workflows/push-deploy.yml
+++ b/.github/workflows/push-deploy.yml
@@ -5,7 +5,7 @@ on:
     types: [published]
 
 permissions:
-  contents: read
+  contents: write
   packages: read
   id-token: write
 


.github/workflows/push-deploy.yml
```diff
@@ -5,7 +5,7 @@ on:
     types: [published]
 
 permissions:
-  contents: read
+  contents: write
   packages: read
```
jeffpaul, line 9: Does this also need to be updated to `write` as done in #345 (and currently underway updating across other plugin repos)?

dkotter: So I know we have this set to `write` in some of our other repos but I don't think it needs to be, though I guess I don't know that for sure and hard to test until we do an actual release. Happy to update that just to be safe if we want

jeffpaul: Maybe better to start with just this change and then can add the `packages: write` update later if the run fails again?

jeffpaul: I'm going to go ahead and update all the other PRs I opened to remove the `packages: write` change and merge those in so we get everything to the same contents:write/packages:read state and can iterate everything together in the future if needed.

jeffpaul: Though I guess also worth clarifying here before I march off and make broad changes, is the `id-token: write` bit here something we should replicate across the other deploy actions?

dkotter: Looked at a few other plugins that have had successful releases (including attaching a release asset) since adding custom permissions. The three I looked at ([Safe SVG](https://github.com/10up/safe-svg/blob/develop/.github/workflows/wordpress-plugin-deploy.yml), [Convert to Blocks](https://github.com/10up/convert-to-blocks/blob/develop/.github/workflows/dotorg-push-deploy.yml), and [Autoshare for Twitter](https://github.com/10up/autoshare-for-twitter/blob/develop/.github/workflows/deploy-to-wpdotorg.yml) all have `contents: write` and `packages: write` but none of them have `id-token: write`, so I don't think we need that

jeffpaul (approved):

PR #365, opened 2026-03-20
feat: add compatibility with WP Offload Media

hegel-berkeley:
### Description of the Change
This PR adds compatibility between `Simple Local Avatars` and `WP Offload Media` (Lite/Pro) so avatar images can be offloaded and served from S3/CDN URLs instead of relying on local uploads URLs.

#### What changed
- Added a dedicated compatibility layer in `includes/class-s3-compatibility.php`.
- Loaded the compatibility file from `simple-local-avatars.php` when Offload Media is detected (`Amazon_S3_And_CloudFront`, `AS3CF_Pro`, or `as3cf_init`).
- Added hooks to:
  - offload avatar attachments after updates (`simple_local_avatar_updated`);
  - resolve avatar URLs through S3/CDN early and late (`pre_simple_local_avatar_url`, `get_avatar_url`, `get_avatar`);
  - swap attachment URLs for avatar-related media (`wp_get_attachment_url`, `wp_get_attachment_image_url`).
- Implemented support for different Offload Media APIs across versions (newer and legacy methods).
- Added logic to:
  - ensure avatar attachments exist in S3;
  - build size-specific avatar URLs when possible;
  - generate missing avatar thumbnails on demand and re-offload;
  - replace `src` and `srcset` in avatar HTML with CDN URLs;
  - detect whether an attachment is used as a user/default avatar.

#### Why
Without this compatibility layer, avatar URLs may still point to local media paths in environments using WP Offload Media. This can cause broken avatars or inconsistent delivery when media is expected to come from S3/CDN.

#### Benefits
- Consistent avatar delivery in offloaded media environments.
- Better compatibility with S3/CDN-based WordPress setups.
- Reduced risk of local URL leakage for avatar images.
- Improved support across WP Offload Media versions.

#### Alternatives / trade-offs considered
- Handling URL replacement only in `get_avatar_url` was considered, but `get_avatar` HTML rewriting was also added to cover additional rendering paths.
- Extra checks/offload attempts may add some overhead, but improve reliability and reduce broken-avatar scenarios.

Closes #

### How to test the Change
1. Install and activate `Simple Local Avatars`.
2. Install and activate `WP Offload Media` (Lite or Pro), configured with a working bucket/CDN.
3. Upload and assign a local avatar to a user profile.
4. Verify in admin/profile and frontend that avatar URLs point to S3/CDN instead of local uploads URLs.
5. Request/render multiple avatar sizes (e.g. 48, 96, 192) and confirm expected size URLs are returned.
6. Confirm avatar HTML output uses CDN URLs in both `src` and `srcset` (where applicable).
7. Change/remove an avatar and verify no regressions in avatar rendering behavior.
8. Verify default avatar behavior still works as expected.
9. Run existing plugin tests and confirm they pass.

### Credits
Props @hegel-berkeley

### Checklist:
- [] I have updated the documentation accordingly.
- [x] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.

### DIFF
--- a/includes/class-s3-compatibility.php
+++ b/includes/class-s3-compatibility.php
@@ -1,0 +1,739 @@
+<?php
+/**
+ * WP Offload Media S3 Compatibility
+ *
+ * This class adds compatibility between Simple Local Avatars and WP Offload Media Lite
+ * to ensure avatars are uploaded to S3 and use the CDN URL.
+ *
+ * @package SimpleLocalAvatars
+ */
+
+/**
+ * S3 Compatibility Class
+ */
+class Simple_Local_Avatars_S3_Compatibility {
+
+	/**
+	 * Instance of this class
+	 *
+	 * @var Simple_Local_Avatars_S3_Compatibility
+	 */
+	private static $instance = null;
+
+	/**
+	 * Get instance of this class
+	 *
+	 * @return Simple_Local_Avatars_S3_Compatibility
+	 */
+	public static function get_instance() {
+		if ( null === self::$instance ) {
+			self::$instance = new self();
+		}
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor
+	 */
+	private function __construct() {
+		$this->init_hooks();
+	}
+
+	/**
+	 * Initialize hooks
+	 */
+	private function init_hooks() {
+		// Hook to ensure avatars are uploaded to S3 after they are assigned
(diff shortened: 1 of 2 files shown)


jeffpaul:
@hegel-berkeley thanks for the PR!  I'll note that @dkotter and I are pretty heavily focused on continued releases on the https://github.com/wordPress/ai plugin for the next couple of weeks and thus likely won't get to this PR for a bit, so some patience waiting for us to be available would be appreciated... thanks again!

PR #370, opened 2026-06-09
fix: validate file type before previewing avatar upload

### LINKED ISSUE #233: Missing Check for File Type Before Uploading Media for Avatar
### Location 1: Default Avatar Setting

**Issue Description:**

#### Any file type can currently be uploaded as an avatar in the file uploader within the Default Avatar Setting. This results in non-useful media being stored in the media library.

**Expected Result:**

The system should verify the file type before uploading it to the media library, ensuring that only supported image formats are accepted.

**Steps to Reproduce:**

1. Navigate to the Default Avatar Setting.
2. Click on the 'Choose Default Avatar' button.
3. Attempt to upload an unsupported file type, such as .zip, .pdf, .heic, etc.
4. Access the media library folder.
5. Observe the uploaded files.

https://github.com/10up/simple-local-avatars/assets/6384299/bb0b690a-9464-4a22-ac55-de42cbe4797b

--------

### Location 2: Avatar Setting Within User Profiles

**Issue Description:**

#### When users upload unsupported file types as their avatars, it results in a broken display on the crop screen area, leading to a suboptimal user experience.

**Steps to Reproduce:**

1. Access your user profile settings.
2. Navigate to the Avatar Setting section.
3. Attempt to upload an avatar file in an unsupported file format (e.g., .zip, .pdf, .heic).
4. After selecting the unsupported file, proceed to the next step, which typically involves cropping or adjusting the avatar.
5. Observe the display on the crop screen area.

**Expected Result:**

The avatar upload process should gracefully handle unsupported file formats, providing an informative message to the user. 

**Actual Result:**

Currently, when an unsupported file type is uploaded as an avatar, it leads to a broken display on the crop screen area within the user profile, impacting the user experience.

https://github.com/10up/simple-local-avatars/assets/6384299/4fd5621f-3198-4e95-9375-1514e76634ac

### Steps to Reproduce

Shared above

Screenshot 2023-09-04 at 1 45 26 PM

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

thisismyurl:
### Description of the Change

The local-file avatar file input accepted any file type — a user could select a PDF, video, or binary and `URL.createObjectURL()` would attempt to render it as an image preview. Two changes close this gap:

- **PHP (`class-simple-local-avatars.php`):** Add `accept="image/jpeg,image/png,image/gif"` to the `<input type="file">` so the OS file picker defaults to images and supporting browsers pre-filter the dialog.
- **JS (`assets/js/simple-local-avatars.js`):** Add a MIME-type guard in the `change` event handler that mirrors the server-side allow-list (`image/jpeg`, `image/gif`, `image/png`). If the selected file's reported `type` is absent or outside that set, the input is cleared, the existing avatar preview is restored, and a localised rejection message is inserted and announced via `wp.a11y.speak()` for screen readers. The `wp-a11y` script is added as a dependency so `wp.a11y` is always available.
- **Tests (`tests/cypress/integration/avatar-file-type.test.js`):** Three Cypress tests cover the rejection, empty-type, and valid-image paths. They run as a subscriber because admins see the media-library uploader rather than the plain file input this feature guards.

Closes #233

### How to test the Change

1. Log in as a user without the `upload_files` capability (e.g. the subscriber created by `tests/bin/initialize.sh`).
2. Go to **Users → Profile**.
3. Under the local avatar field, attempt to select a non-image file (e.g. `.pdf`, `.txt`, `.exe`).
4. Confirm the input is cleared, the existing avatar preview is unchanged, and an error message appears below the field.
5. Confirm selecting a valid JPEG, PNG, or GIF image clears the error and updates the preview.
6. Confirm the error message is announced by a screen reader (if available).
7. Confirm the admin/media-library upload path is unaffected.

### Changelog Entry

- Add `accept` attribute to avatar file input to filter the OS file picker to images
- Add client-side MIME-type guard that rejects non-image files before preview and surfaces an accessible error message

### Credits

Props @thisismyurl

### DIFF
--- a/assets/js/simple-local-avatars.js
+++ b/assets/js/simple-local-avatars.js
@@ -147,6 +147,44 @@ jQuery(document).ready(function ($) {
 	});
 
 	/**
+	 * Image MIME types the server will actually accept on upload. Mirrors the
+	 * `mimes` allow-list passed to media_handle_upload() in
+	 * Simple_Local_Avatars::edit_user_profile_update(), so the client never
+	 * previews a file the server would reject on submit.
+	 */
+	const avatar_allowed_types = ['image/jpeg', 'image/gif', 'image/png'];
+
+	/**
+	 * Show a rejection message for an invalid avatar file and announce it to
+	 * assistive technology.
+	 *
+	 * @param {string} message Human-readable reason the file was rejected.
+	 */
+	function avatar_show_error(message) {
+		let $error = $('#simple-local-avatar-error');
+		if (!$error.length) {
+			$error = $('<span/>', {
+				id: 'simple-local-avatar-error',
+				class: 'description',
+				css: { display: 'block', color: '#d63638' },
+			});
+			avatar_input.after($error);
+		}
+		$error.text(message);
+
+		if (window.wp && wp.a11y && 'function' === typeof wp.a11y.speak) {
+			wp.a11y.speak(message, 'assertive');
+		}
+	}
+
+	/**
+	 * Clear any avatar rejection message.
+	 */
+	function avatar_clear_error() {
+		$('#simple-local-avatar-error').remove();
+	}
+
+	/**
 	 * Update the Local Avatar image
 	 */
 	avatar_input.on('change', function (event) {
@@ -155,14 +193,23 @@ jQuery(document).ready(function ($) {
 		URL.revokeObjectURL(avatar_blob);
 		if (event.target.files.length > 0) {
 			const file = event.target.files[0];
-			if (file.type && 0 !== file.type.indexOf('image/')) {
+
+			// Reject anything without a MIME type (e.g. a renamed binary the
+			// browser cannot identify, where file.type is an empty string) or
+			// outside the server's accepted set. indexOf() returns -1 for an
+			// empty/unknown type, so this guard never previews an unsafe file.
+			if (-1 === avatar_allowed_types.indexOf(file.type)) {
 				avatar_input.val('');
 				avatar_preview.attr('src', current_avatar);
+				avatar_show_error(i10n_SimpleLocalAvatars.invalidFileType);
 				return;
 			}
+
+			avatar_clear_error();
 			avatar_blob = URL.createObjectURL(file);
 			avatar_preview.attr('src', avatar_blob);
 		} else {
+			avatar_clear_error();
 			avatar_preview.attr('src', current_avatar);
 		}
 	});
--- a/assets/js/simple-local-avatars.js
+++ b/assets/js/simple-local-avatars.js
@@ -194,10 +194,12 @@ jQuery(document).ready(function ($) {
 		if (event.target.files.length > 0) {
 			const file = event.target.files[0];
 
-			// Reject anything without a MIME type (e.g. a renamed binary the
-			// browser cannot identify, where file.type is an empty string) or
-			// outside the server's accepted set. indexOf() returns -1 for an
-			// empty/unknown type, so this guard never previews an unsafe file.
+			/*
+			 * Reject anything without a MIME type (e.g. a renamed binary the
+			 * browser cannot identify, where file.type is an empty string) or
+			 * outside the server's accepted set. indexOf() returns -1 for an
+			 * empty/unknown type, so this guard never previews an unsafe file.
+			 */
 			if (-1 === avatar_allowed_types.indexOf(file.type)) {
 				avatar_input.val('');
 				avatar_preview.attr('src', current_avatar);
--- a/assets/js/simple-local-avatars.js
+++ b/assets/js/simple-local-avatars.js
@@ -154,7 +154,13 @@ jQuery(document).ready(function ($) {
 		avatar_preview.attr('height', 'auto');
 		URL.revokeObjectURL(avatar_blob);
 		if (event.target.files.length > 0) {
-			avatar_blob = URL.createObjectURL(event.target.files[0]);
+			const file = event.target.files[0];
+			if (file.type && 0 !== file.type.indexOf('image/')) {
+				avatar_input.val('');
+				avatar_preview.attr('src', current_avatar);
+				return;
+			}
+			avatar_blob = URL.createObjectURL(file);
 			avatar_preview.attr('src', avatar_blob);
 		} else {
 			avatar_preview.attr('src', current_avatar);
--- a/includes/class-simple-local-avatars.php
+++ b/includes/class-simple-local-avatars.php
@@ -846,7 +846,7 @@ class Simple_Local_Avatars {
 
 		$this->remove_nonce = wp_create_nonce( 'remove_simple_local_avatar_nonce' );
 
-		wp_enqueue_script( 'simple-local-avatars', plugins_url( '', dirname( __FILE__ ) ) . '/dist/simple-local-avatars.js', array( 'jquery' ), SLA_VERSION, true );
+		wp_enqueue_script( 'simple-local-avatars', plugins_url( '', dirname( __FILE__ ) ) . '/dist/simple-local-avatars.js', array( 'jquery', 'wp-a11y' ), SLA_VERSION, true );
 		wp_localize_script(
 			'simple-local-avatars',
 			'i10n_SimpleLocalAvatars',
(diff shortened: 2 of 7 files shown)


assets/js/simple-local-avatars.js
```diff
@@ -154,7 +154,13 @@ jQuery(document).ready(function ($) {
 		avatar_preview.attr('height', 'auto');
 		URL.revokeObjectURL(avatar_blob);
 		if (event.target.files.length > 0) {
-			avatar_blob = URL.createObjectURL(event.target.files[0]);
+			const file = event.target.files[0];
+			if (file.type && 0 !== file.type.indexOf('image/')) {
```
thrijith: `if (file.type && 0 !== file.type.indexOf('image/'))`. The leading `file.type &&` skips the guard when `file.type` is `''` (some browsers/OS for unrecognized types, e.g. a renamed binary), so `createObjectURL` runs on an unknown file. 

Should the guard treat empty type as a reject? 

Also as suggested by Copilot AI `if (!file.type || !file.type.startsWith('image/'))` also reads cleaner than `indexOf`.

assets/js/simple-local-avatars.js
```diff
@@ -154,7 +154,13 @@ jQuery(document).ready(function ($) {
...
+				avatar_input.val('');
+				avatar_preview.attr('src', current_avatar);
+				return;
+			}
```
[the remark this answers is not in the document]
thrijith: Same as above, on reject the selection just disappears with no message. Verified on the profile page: after selecting a `.txt`, input cleared (0 files), zero notices added, section text unchanged, no alert. Should it surface a brief inline notice so the user knows why?

includes/class-simple-local-avatars.php
```diff
@@ -990,7 +990,7 @@ public function edit_user_profile( $profileuser ) {
 										?>
 										<p style="display: inline-block; width: 26em;">
 											<span class="description"><?php esc_html_e( 'Choose an image from your computer:', 'simple-local-avatars' ); ?></span><br />
-											<input type="file" name="simple-local-avatar" id="simple-local-avatar" class="standard-text" />
+											<input type="file" name="simple-local-avatar" id="simple-local-avatar" class="standard-text" accept="image/*" />
```
thrijith: Server accepts only jpeg/gif/png https://github.com/10up/simple-local-avatars/blob/a26851f1b0b3fb082fd0aec1e48dd1b2eae01658/includes/class-simple-local-avatars.php#L1132-L1136. 

The JS `image/*` guard passes svg/webp, which preview then fail on submit. Should the JS check mirror the server set (or at least exclude `image/svg+xml`) for consistent UX?

thrijith (changes requested):

thisismyurl:
Thanks @thrijith — all three addressed in the latest push.

**1. Empty / unknown MIME no longer slips through.** The old `if (file.type && 0 !== file.type.indexOf('image/'))` short-circuited when `file.type` was `''` (the renamed-binary case you flagged), so `createObjectURL` ran on an unknown file. The guard is now an allowlist check:

```js
const avatar_allowed_types = ['image/jpeg', 'image/gif', 'image/png'];
// indexOf('') === -1, so empty/unknown types are rejected, not previewed.
if (-1 === avatar_allowed_types.indexOf(file.type)) { ... }
```

I went with an explicit allowlist rather than your suggested `!file.type || !file.type.startsWith('image/')` for one reason: `startsWith('image/')` would still admit `image/svg+xml` and `image/webp`, which is exactly the preview-then-fail gap in point 3. The allowlist closes that in the same check.

**2. Client now mirrors the server's accepted set.** The server only accepts jpeg/gif/png (the `mimes` array on `media_handle_upload()` in `edit_user_profile_update()`), so the JS uses the same three, and I tightened the input to `accept="image/jpeg,image/png,image/gif"` so the native picker filters too. svg/webp no longer preview only to fail on submit.

**3. Rejections are surfaced, not silent.** On reject the input now shows a visible inline notice (`#simple-local-avatar-error`) and announces via `wp.a11y.speak( message, 'assertive' )`. I added `wp-a11y` as a script dependency and feature-guard the call (`window.wp && wp.a11y && typeof wp.a11y.speak === 'function'`) so it degrades quietly if the dependency is ever absent. The assertive politeness level is intentional — a rejected upload is a now-relevant correction, not a passive status.

Added `tests/cypress/integration/avatar-file-type.test.js` covering the text-file reject, the empty-MIME reject, and the valid-image happy path.

One note on verification: I ran `php -l` and a JS syntax check locally, but I wasn't able to spin up `wp-env`/Cypress or `npm run lint:js` in my environment, so the new e2e specs and lint will get their first real run in CI on this push — I'll keep an eye on the checks and follow up here if anything needs a touch-up.

(full disclosure: AI helped me identify the issue and verify my work)

thrijith:
Hi @thisismyurl 

Thanks a lot for the updates, code looks good to me.

> Added tests/cypress/integration/avatar-file-type.test.js covering the text-file reject, the empty-MIME reject, and the valid-image happy path.

The tests seem to be failing because it is logging in as admin who won't see the file input control, can you set up a user with subscriber role in tests see if that helps?

thisismyurl:
Done, thanks for the review

thrijith:
Passing to @peterwilsoncc for final review the last failing test is unrelated to the changes done here, please let me know if that should handled in this PR, thanks!

peterwilsoncc:
> Passing to @peterwilsoncc for final review the last failing test is unrelated to the changes done here, please let me know if that should handled in this PR, thanks!

I've pushed https://github.com/10up/simple-local-avatars/pull/370/commits/655727ee834109eaa8b568c50ce11b74c4de843b to resolve by navigating to the dashboard as `wpApiSettings` are defined there.

thisismyurl:
Thanks @peterwilsoncc, that did it. Pulling `wpApiSettings.nonce` from the dashboard was the missing piece, and the Cypress suite is green now.

The one remaining red is `dependency-review`, and it looks unrelated to this PR. It fails with:

> Unable to fetch or parse config file: Error fetching remote config file

The workflow points at an org-level config, `config-file: 10up/.github/.github/dependency-review-config.yml@trunk`, and the step is failing to fetch it rather than flagging anything in the diff. This PR touches no dependency manifests (just the avatar-upload JS and the Cypress test), and the action runs with `vulnerability-check: false`, so there's nothing here for it to catch either way. Reads like a transient or access issue fetching that shared config, not something in this branch.

Happy to rebase or adjust anything if it would help, but I don't think there's a code change on my side that clears that check. Thanks again to you and @thrijith for carrying it through review.

peterwilsoncc:
@thisismyurl

The failing dependency review test is now passing, it was probably a permissions error or similar as you're not in the `@10up` organisation

For users with upload permissions (ie, authors and above) it would be good if the same limitation were applied to the media library upload tab. In the block editor's image block, all users are limited to the uploading of images only but I'm not sure if the same limitation can be applied to the media modal elsewhere, are you willing to look in to this?

I don't think it's a blocker for merge and can be done in a follow up if you can't configure it easily. The WP media modal is a complex beast.

assets/js/simple-local-avatars.js
```diff
@@ -146,6 +146,44 @@ jQuery(document).ready(function ($) {
 			});
 	});
 
+	/**
+	 * Image MIME types the server will actually accept on upload. Mirrors the
+	 * `mimes` allow-list passed to media_handle_upload() in
+	 * Simple_Local_Avatars::edit_user_profile_update(), so the client never
+	 * previews a file the server would reject on submit.
+	 */
```
peterwilsoncc, line 154:
```suggestion
	 *
	 * @var {Array} Permitted image types.
	 */
```

assets/js/simple-local-avatars.js
```diff
@@ -154,9 +192,24 @@ jQuery(document).ready(function ($) {
 		avatar_preview.attr('height', 'auto');
 		URL.revokeObjectURL(avatar_blob);
 		if (event.target.files.length > 0) {
-			avatar_blob = URL.createObjectURL(event.target.files[0]);
+			const file = event.target.files[0];
+
+			// Reject anything without a MIME type (e.g. a renamed binary the
+			// browser cannot identify, where file.type is an empty string) or
+			// outside the server's accepted set. indexOf() returns -1 for an
+			// empty/unknown type, so this guard never previews an unsafe file.
```
peterwilsoncc: Nit: The WPCS for inline docs are to use the following as a multiline comment format.

```suggestion
			/*
			 * Reject anything without a MIME type (e.g. a renamed binary the
			 * browser cannot identify, where file.type is an empty string) or
			 * outside the server's accepted set. indexOf() returns -1 for an
			 * empty/unknown type, so this guard never previews an unsafe file.
			 */
```

peterwilsoncc (changes requested):
A couple of minor nitpicks inline.

jeffpaul:
@thisismyurl are you able to update based on last code review feedback?

thisismyurl:
Pushed 8bddfb7 — addresses both remaining nits.

The multiline-comment WPCS format was already applied in c97a290 (Address review feedback on #370), so the only outstanding item was the JSDoc block on `avatar_allowed_types`. I used `@member {Array} Permitted image types.` rather than the literal `@var` suggestion, since this repo's own eslint config (`jsdoc/check-tag-names`) flags `@var` in favor of `@member` — ran `npm run lint:js` locally to confirm the swap resolves clean with no new warnings introduced.

@jeffpaul this should be ready for another look.

jeffpaul:
_Awaiting results of workflow runs..._

